PR #12, opened 2016-01-13
Remove es6 syntax that required babel deps

### LINKED ISSUE #10: Object assign error
Hey,

I'm importing mapbox-gl-geocoder from npm and using it with browserify. I get the following error:

`Unknown plugin "object-assign" while parsing file: .../node_modules/mapbox-gl-geocoder/index.js`

Any clue ?

EDIT : 
Got it to work with https://www.npmjs.com/package/babel-plugin-object-assign since I'm using babelify, but I found another error:

`geocoder.js:6 Uncaught ReferenceError: mapboxgl is not defined` which is this line : 

`export default class Geocoder extends mapboxgl.Control {`

My code is the following
```
import mapboxgl             from 'mapbox-gl'; // 0.12.1
import Geocoder             from 'mapbox-gl-geocoder';
```

tristen:
If this plugin is added to a project via npm the project had to also respect
ES6 syntax by using a compiler like Babel. This is expecting too much.

Other notable changes:

- Use envify inline with the npm test command
- Use xtend module over Object.assign. Closes #10

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,0 @@
-{
-  "stage": 0,
-  "loose": "all",
-  "plugins": ["object-assign"]
-}
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,4 @@
 {
-  "parser": "babel-eslint",
   "env": {
     "node": true,
     "browser": true
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@
  * map.addControl(geocoder);
  * @return {Geocoder} `this`
  */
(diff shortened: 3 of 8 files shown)


tristen:
Pointing [`main`](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/package.json#L5) to [dist/mapbox-gl-geocoder.js](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/dist/mapbox-gl-geocoder.js) should also work :p

PR #22, opened 2016-02-10
Babel 6

drewbo:
Upgrades to the Babel 6 presets system. Also bumps the `mapbox-gl` peerDependencies (somewhat arbitrarily chose those of `mapbox-gl-draw`).

Tests pass locally but happy to do more work here to get it ready for primetime


tristen:
Thanks @drewbo!

PR #26, opened 2016-02-29
Bug 25 ie styling

### LINKED ISSUE #25: Incorrect styling in IE
screen shot 2016-02-29 at 1 57 52 pm

Tested with Browserstack, Windows 10, IE 11

drewbo:
Close #25 

I cleaned up the icon source (though that turned out not to be the issue, at least they are slightly smaller now?)

Turns out the IE ignores all the arguments to `classList.add` except the first


drewbo:
Good call, updated

PR #31, opened 2016-03-02
Update the mapbox-gl peerDependency requirement

drewbo:
At some point we can probably leave old `mapbox-gl` versions behind but I figured this was safer for now


tristen:
Hah .. yes we probably should. :+1: for now!

PR #46, opened 2016-06-28
Replace request with isomorphic-fetch

sleepycat:
This change started it's life as a bughunt.

While debugging the error `A wildcard '*' cannot be used in the 'Access-Control-Allow-Origin'
header when the credentials flag is true` I came across
request/request#986. markbahnman's comment in that issue resolves that
particular error which then revealed another request related error
`TypeError: self.req.abort is not a function`.

Given the characterization of request as "a Node.js library,
used server-side" and the errors, it seemed an opportune moment to
explore removing request and exploring another library.

This commit does exactly that, removing `request` and replacing it with
`isomorphic-fetch`.

One open question with this approach is that the old code would cancel
this.request if it existed. Cancelling a fetch is under discussion in
whatwg/fetch#27 is as yet unresolved.

Apart from that, the tests all pass, the lint is clean, and it's working in my project so I thought I would send a PR for your consideration.
If there are other changes/improvements that need to be made, let me know.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,8 @@ if (!mapboxgl) throw new Error('include mapboxgl before mapbox-gl-geocoder.js');
 var Typeahead = require('suggestions');
 var debounce = require('lodash.debounce');
 var extend = require('xtend');
-var request = require('request');
+require('es6-promise').polyfill();
+var fetch = require('isomorphic-fetch');
 var EventEmitter = require('events').EventEmitter;
 
 // Mapbox Geocoder version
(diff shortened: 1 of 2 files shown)


tristen:
Heya @sleepycat! Thanks for the PR. I missed this when I committed https://github.com/mapbox/mapbox-gl-geocoder/commit/81919f389b3a49f16b86adc513d23a353ea3cb8b yesterday. Can you confirm the issue is resolved using master?

sleepycat:
Hi @tristen I tried out master and it seems to be resolved. That's funny that we were both thinking the same thing at the same time. Glad to see it fixed!

tristen:
@sleepycat glad to hear! request was actually blowing up the file size of the minified `dist/mapbox-gl-geocoder.js` which is why I opted for plain old `xmlhttprequest`.

PR #50, opened 2016-07-20
minor jsdoc change

andrewharvey:
Since these two values just a one dimensional array, shouldn't the type be just Array<number> ?


tristen:
👍

PR #58, opened 2016-09-15
preventDefault on clear

### LINKED ISSUE #56: Form submitted on input clear when geocoder is within form
When the geocoder div is within a form and `.geocoder-icon-close` button clicked or user hits return after selecting a suggestion from keyboard, the form is submitted.

I confirmed this by taking the example [code from here](https://www.mapbox.com/mapbox-gl-js/example/point-from-geocoder-result/) and adding a wrapping form around `div.map` and `div.geocoder-container`. Then I entered an address with suggestions and clicked the "x" (aka `button.geocoder-icon-close`). The form was submitted rather than simply clearing the input value.

I then also repeated the test, but instead of clicking the "x", I arrowed down to select a suggestion and clicked return. The form was also submitted.

This was in Safari, Chrome, and Firefox.

The mapbox-gl-geocoder is really handy! Thanks for all your great work. Just a little detail that would be great to fix or have insight about how to eliminate the behavior.

tristen:
Fixes #56 @walter can you confirm?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -174,7 +174,8 @@ MapboxGeocoder.prototype = {
     return request;
   },
 
-  _clear: function() {
+  _clear: function(ev) {
+    if (ev) ev.preventDefault();
     this._inputEl.value = '';
     this._typeahead.selected = null;
     this._typeahead.clear();


walter:
Almost!

It prevents the form from being submitted when clicking the "X" to clear the input and it also does not  submit the form when using the arrows and returns to select a value. Great.

However, when you use the arrows and return to select the value, although it updates the map, it doesn't update the input value to match. It's just blank.

walter:
Took a look to see if I could figure anything out. 🤔 Didn't see anything obvious unfortunately.

tristen:
@walter thanks! it looks like this only half solves things :/ there's an issue in [Suggestions](https://github.com/tristen/suggestions) where mousedown and keyboard returns are handled differently. Just need to set some time aside to fix it.

walter:
A 👍  is not quite the right response. Thanks for working this! I understand these things can take time.

andrewharvey:
I'm bitten by this bug too. I tried debugging why the input clears when selecting a value via the keyboard, I noticed that MapboxGeocoder._clear was being called with a `click` event on the X button when using the keyboard to select a value from the list.

Unfortunately I can't work out why this click event is being erroneously emitted.

andrewharvey:
Based on my testing this PR together with https://github.com/tristen/suggestions/issues/15 fixes the form submission issue.

PR #60, opened 2016-10-06
Allows module to be imported/required

richsilv:
At present, if the `mapbox-gl` module has been imported/required rather than added to global scope, an error will be thrown when you try to import/require this module.  This makes it impossible to use with something like [react-map-gl](https://github.com/uber/react-map-gl).

This PR removes the check and replaces it with code which either assigns either the global `mapboxgl` instance, or a required version of the module, to the local `mapboxgl`, allowing it to be used in either case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,11 @@
 'use strict';
 
-/* global mapboxgl */
-if (typeof mapboxgl === 'undefined') throw new Error('include mapboxgl before mapbox-gl-geocoder.js');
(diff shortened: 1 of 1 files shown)


tristen:
This would include the mapbox-gl source in browserify builds always ☹️ there's a convo happening in https://github.com/mapbox/mapbox-gl-js/issues/1835 that should supersede the state of things.

Issue #59, opened 2016-10-04
Weird search results for United States, Canada and Russia


magnus-eriksson:
When searching for United States, Canada or Russia, the map zooms out, showing the whole world instead of zooming in on those particular countries. It looks like it tries to do some positioning vertically, but it's not really accurate.

We have a map where we show stores all over the world, including in these three countries. so it's quite important that these, large, countries gets handled correctly.

lbud:
Thanks for your report @magnus-eriksson. This is indeed unexpected behavior from a user's POV. What's happening is that mapbox-gl-geocoder zooms to the bounding box returned from the geocoding API, which includes the legal entirety of a country (so in the case of the USA the Aleutian Islands cross the International Date Line, or France includes the island of Reunion in the Indian Ocean, etc…)

The problem with bounding boxes including outlying islands is a complicated and political one and may be addressed in our geocoder eventually but isn't really on the roadmap right now. In the short run I'm addressing with #66.

PR #66, opened 2016-11-02
Provide exceptions to country bboxes when using `flyTo`

lbud:
Fixes https://github.com/mapbox/mapbox-gl-geocoder/issues/59.

As pointed out in that issue, selecting some countries shows a much larger swath of the world than expected. This is because we fly to the result's geocoder-provided `bbox`, which encapsulates the entirety of that country's legal boundaries (e.g. "France" zooms to a bbox that includes the island of Reunion; "United States"'s bbox spans the entire world since some of the Aleutian Islands cross the International Date Line).

As a short-term fix here I'm adding an `exceptions` file with (<WARN>totally subjective, I made these up</warn>) "reasonable" zoom levels, so when any of the countries therein are selected, rather than zooming to the result's `bbox`, it will use the result's `center` in combination with the hardcoded zoom level.

cc @tristen @sbma44 @tmcw @magnus-eriksson


tmcw:
Definitely in favor of providing a more conventional result for searching for 'France'. On a technical level, I think zoom levels are a little too rough for this - a zoom + centerpoint combination is functionally equivalent to a bbox, but unlike a bbox, these hardcoded zoom levels will look way off on tiny and large maps, and only appropriate on medium-ish sized maps.

lbud:
Rats @tmcw you're right. I hoped to avoid the political trap that is defining selective bboxes here, but that might be our best bet. 😕

Issue #67, opened 2016-11-10
Use new IControl interface for control registration


tmcw:
https://github.com/mapbox/mapbox-gl-js/pull/3497

The new IControl interface will allow mapbox-gl-geocoder to remove any dependency on mapboxgl, eliminating a popular class of installation bugs. Once mapboxgl is updated, or before if you're feeling saucy, update to the new interface.

tmcw:
Fixed in https://github.com/mapbox/mapbox-gl-geocoder/pull/70

PR #70, opened 2016-11-13
Fix compatibility with 0.27

tmcw:
Fixes

* #67 - makes MapboxGeocoder compatible with new control system
* #71 - no longer relies on a bundler to bundle JSON
* #72 - no longer allows [lng, lat] input. Users need to call `.join()` themselves.
* #74 - no longer exposes `fire` as a public API method.
* #75 - remove indirection
* #76 - remove event creation in order to call internal methods
* #77  - remove `getResult` method that returned the input
* #79 - remove nonsense pattern in tests
* #80 - no longer renames 'features' to 'results' in `results` event.
* #83 - fixes test .pass calls


tmcw:
Okay, I think this is ready for review. It fails on Circle because there's no WebGL, but it passes tests and works with `npm start` locally. cc @tristen for the review

index.js
```diff
@@ -58,12 +43,6 @@ Geocoder.prototype = inherit(mapboxgl.Control, {
   onAdd: function(map) {
     this.request = new XMLHttpRequest();
 
-    this.container = this.options.container ?
-      typeof this.options.container === 'string' ?
-      document.getElementById(this.options.container) :
-      this.options.container :
-      map.getContainer();
```
tmcw: Because the PR is about updating to the 0.27.0 API, and the API defines that you return an element and the map adds that element. Attaching to an arbitrary DOM node is not part of the API.

tristen: The ability to attach the geocoder to a DOM node outside of the map container is important.

tmcw: It might be important, but as of right now it's impossible. You'll need to lobby for that as an API feature on the GL JS side, while this PR will fix the arguably more important issue of this control being broken.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+### v2.0.0
+
+- Support for the Mapbox GL JS 0.27.0 API
+
+Breaking changes:
+
+- `setInput` and `query` methods no longer accept a `[lng, lat]` array. If you'd
+  like to search for a location and you have that data as `[lng, lat]`, call
+  `.join()` on the array before passing it to the geocoder control.
+- `.fire` method removed
```
tmcw: Added

tristen:
👍

Issue #98, opened 2017-02-15
API call made after 1 character typed even though no results shown until 2 characters typed


andrewharvey:
[Inherited from Suggestions](https://github.com/tristen/suggestions/blob/gh-pages/API.md) by default two characters need to be typed into an input to trigger suggestions.

However the Geocoder still makes an API call to the Geocoding API after 1 character is typed. This wastes Geocoding API quota since it doesn't seem to be used anywhere.

Is this right? Would you accept a PR to skip the API call when the result won't be displayed?

tristen:
https://github.com/mapbox/mapbox-gl-geocoder/pull/103

PR #103, opened 2017-02-23
Add limit and minLength options. Avoid unnecessary Geocode calls after 1 character. Fixes #97 #98

### LINKED ISSUE #97: expose Suggestions limit and minLength as options for the Geocoder
Would you accept a PR adding the two [Suggestions parameters](https://github.com/tristen/suggestions/blob/gh-pages/API.md) limit and minLength as options for the Geocoder?

    -   `options.limit` **[Number]** Max number of results to display in the auto suggest list.
         (optional, default `5`)
    -   `options.minLength` **[Number]** Number of characters typed into an input to trigger suggestions.
         (optional, default `2`)

andrewharvey:
the `geocode` method called will still apply even when length is < minLength. I think that's the intention of geocode. The minLength only applies to users typing in the input.

Addresses #97 and #98.

### DIFF
--- a/API.md
+++ b/API.md
@@ -23,6 +23,8 @@ A geocoder component using Mapbox Geocoding API
         for available types.
     -   `options.country` **\[[string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String)]** a comma separated list of country codes to
         limit results to specified country or countries.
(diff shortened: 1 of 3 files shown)


tristen:
🎉 Looks like this commit includes both https://github.com/mapbox/mapbox-gl-geocoder/pull/101 & https://github.com/mapbox/mapbox-gl-geocoder/pull/102. Do you mind pulling those out?

andrewharvey:
I've rebased it which seems to have fixed it that, (only the one commit showing up in this branch now)

PR #107, opened 2017-02-25
aria-label on the Clear button

andrewharvey:
working on further changes upstream at https://github.com/tristen/suggestions/pull/14 which might require more changes here


tristen:
Looks good. Is this still a WIP?

andrewharvey:
This on it's own is ready to merge. There could be further changes to make to mapbox-gl-geocoder in light of tristen/suggestions/pull/14 but on second thought, let's merge this and I'll open another PR here if I need to later.

PR #119, opened 2017-05-30
Possible style fix

SKalt:
I closed the `code` markup around the geocoder event data.  If this everything-after-the-event-argument-is-code style was intentional, the **error** tag should also have had it.


andrewharvey:
API.md is automatically generated, you'll need to change the source and run `npm run docs` to update API.md.

SKalt:
Got it.

Issue #118, opened 2017-05-24
Proximity updates based on map movement


keyofj:
When the MapboxGeocoder is initialized we can set the proximity based on the current map center point. It would be nice to be able to have this updated based on where the user has the map when they do subsequent searches.  Not automatic though, I think it should be set as needed to not interfere with specific user experiences.

In our case if someone returns to a saved map they were working on in Dallas, the initial proximity works well if they want to search around Texas. If they jump over to California to work in the map, search results would be returned relative to that current area.

andrewharvey:
what about a `geocoder.setProximity(proximity)` method which would then allow this to be implemented in the application code with:

```
map.on('moveend', function() {
   geocoder.setProximity(map.getCenter());
})
```

tristen:
> what about a `geocoder.setProximity(proximity)` method 

That sounds like a good addition to add.

andrewharvey:
I've merged #122 which now let's you implement this with:

```js
map.on('load', updateGeocoderProximity);
map.on('moveend', updateGeocoderProximity);

function updateGeocoderProximity() {
    if (map.getZoom() > 9) { // proximity is designed for local scale, if the user is looking at the whole world, it doesn't make sense to factor in the arbitrary centre of the map
        var center = map.getCenter().wrap();
        geocoder.setProximity({ longitude: center.lng, latitude: center.lat });
    } else {
        geocoder.setProximity(null);
    }
}

```

I'm leaving this issue open to track automatically doing this by passing a `trackProximity: true` option.

siddmegadeth:
but how to update geocoder search box on mapmove event

andrewharvey:
> but how to update geocoder search box on mapmove event

@siddmegadeth could you elaborate?

PR #122, opened 2017-06-02
set/get proximity methods

andrewharvey:
helps users implement #118 by making the proximity option publicly setable/getable


andrewharvey:
Due to a few minor annoyances to do #118 with this PR you need

```
map.on('moveend', function() {
    var center = map.getCenter().wrap();
    geocoder.setProximity({ longitude: center.lng, latitude: center.lat });
});
```

because:
1. older gl js versions didn't wrap to world 0, and the geocoding api doesn't accept unwrapped coordinates, and neither mapbox-gl-geocoder or mapbox-gl-sdk wrap coordinates
2. `map.getCenter()` returns `{lat: 0, lon: 0}` but mapbox-gl-sdk wants `{longitude: 0, latitude: 0}`

PR #124, opened 2017-07-28
Fix "TypeError: Attempted to assign to readonly property" in Safari 9 on `npm start`

mblomdahl:
Safari 9 fails to open localhost:9966 after `npm start`, console says 
> TypeError: Attempted to assign to readonly property.

about example/index.js style assignments. This fixes it + removes some static code analysis warnings in WebStorm. :)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,5 +3,9 @@
     "node": true,
     "browser": true
   },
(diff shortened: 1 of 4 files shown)


andrewharvey:
Looks good to me.

PR #125, opened 2017-08-01
Fix 'clear' event placement in API doc

snkashis:
Was not part of list with others.

### DIFF
--- a/API.md
+++ b/API.md
@@ -87,7 +87,8 @@ Subscribe to events that happen within the plugin.
 
 **Parameters**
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
@snkashis The API doc is automatically generated with https://github.com/mapbox/mapbox-gl-geocoder/blob/master/package.json#L10 so this change would be overwritten next time the doc is updated. This happens due to an upstream bug https://github.com/documentationjs/documentation/issues/676.

I'm not sure if we should leave it as is and wait for the upstream fix, or merge this and try to manually fix it up every time the doc is generated.

andrewharvey:
Sorry @snkashis but I'm going to close this as I don't think manually fixing this up every time we regenerate the API.md is viable, I'd rather see https://github.com/documentationjs/documentation/issues/676 solved or some other solution which doesn't involve manually tweaking API.md.

PR #123, opened 2017-07-07
[wip] sketch of local geocoder option

andrewharvey:
Not ready to be merged, but creating a PR for visibility. See #121 for background.

Issue #121, opened 2017-06-02
Geocoder does not handle lat/long input


ArstanWhitebeard:
In Google Maps you can type either a lat/long or a geocoder search into the search box and it will fly there. mapbox-gl-geocoder cannot handle the former, which is both surprising to the user and a limit to its usefulness. 

Ideally if you entered a lat/long input it would recognise it (regex?) and simply fly straight there without needing to call to the geocoder api.

andrewharvey:
when one number is >90 or <90 it's obvious which order the coordinates are in, but if both are within -90 to 90 how would you work out if it's lat,lon or lon,lat?

I wonder if it's better to address this upstream in the geocoding api so requests for decimal degree coordinates, dms coordinates work in the api?

andrewharvey:
Currently if you pass in a `lng,lat` it will reverse geocode since this is how the Mapbox Geocoding API https://www.mapbox.com/api-documentation/#request-format. Though with some combinations of `limit` and `types` it can actually return an error.

Putting aside the question of if this should be handled in the upstream API, I've sketched out how this could work in mapbox-gl-geocoder at https://github.com/andrewharvey/mapbox-gl-geocoder/tree/local-geocoder.

The idea is to allow the app developer to supplier their own local geocoder function to supplement results from the external api. In the sketch I've put together a local coordinate pair example, though the beauty is it's flexible so could also be used in other applications where you may have a local GeoJSON source which you want to include in the geocoder results.

This code is just a sketch, before getting it into something which could be merged, I'd like feedback on what others feel about this? 

/cc @tristen

andrewharvey:
https://github.com/mapbox/mapbox-gl-geocoder/blob/master/example/index.js now contains an example of how to do this. I'm thinking it's probably best to leave this up to the implementor (since now with the localGeocoder option it's very easy) rather than including it in core, though I'm not 100% committed to that view.

andrewharvey:
The example of how you can implement this in your own implementation is at https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/. This ticket can be used to track doing this out of the box.

andrewharvey:
[...]

PR #136, opened 2017-12-15
add local geocoder option

andrewharvey:
The idea of this PR is to allow the app developer to supplier their own local geocoder function to supplement results from the external api. Two main use cases are:

- accepting coordinates as input to the geocoder (I've included this in the example)
- providing results from a local GeoJSON dataset

This PR stops short of providing this asynchronously, so you can't call a 3rd party remote API to supplement results. That could be added in a future PR.

replaces #123, helps to support #121 

I'm happy with this now, any feedback welcome.


andrewharvey:
I've had this out since July for review and haven't heard much so given I feel this PR

- includes documentation
- includes tests
- has been in use by myself for a while without noticing any issues
-  doesn't add much weight/bloat
- seems to be useful

I'll merge this PR now, but still welcome any feedback.

anthony-toezin:
I just came across this and agree this is very useful.

andrewharvey:
I'm keen to see this option support an asynchronous function too, you're welcome to submit a PR for this., In the meantime something like https://github.com/ybogdanov/node-sync might work.

PR #151, opened 2018-05-11
add trackProximity option

### LINKED ISSUE #118: Proximity updates based on map movement

andrewharvey:
closes #118 

- [ ] when new version is released, update https://github.com/mapbox/mapbox-gl-js/blob/mb-pages/docs/pages/example/mapbox-gl-geocoder-proximity-bias.html

### DIFF
--- a/API.md
+++ b/API.md
@@ -25,6 +25,7 @@ A geocoder component using Mapbox Geocoding API
         a geographical point given as an object with latitude and longitude
         properties. Search results closer to this point will be given
         higher priority.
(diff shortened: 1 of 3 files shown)


andrewharvey:
@tristen I think you're right about travis being some [org change](https://github.com/mapbox/mapbox-gl-accessibility/pull/16#issuecomment-387727718), this repo uses circleci, but recently it's attempting to use travisci which fails as there is no config.

PR #152, opened 2018-05-14
fix unit tests

andrewharvey:
CircleCI only run's lint, the unit tests must be run locally with `npm run test`.

- updates lib/exceptions with current feature id's returned by the geocoder
- fixes cases where response.features doesn't exist due to a non 200 response from the geocoder, in those cases we still want results from the localGeocoder to be included. This change is needed to fix the broken tests.
- fix localGeocoder unit test, per jsdoc it must return an array of features
- geocoder.once isn't a function only geocoder.on and geocoder.off, however they are wrapped up in once so geocoder.on will suffice.
- other changes to the geocoding api responses hardcoded in the unit tests
- reverse geocoding with limit needs a types option


tristen:
Yikes 🙈 thanks for fixing that up @andrewharvey. 

> CircleCI only run's lint

Also yikes 🙈Ideally Circle would through unit tests too. If I'm following (or remembering) correctly why they can't be run is the test dependency on `smokestack`? CircleCI can't run GL in a headless environment or something?

Issue #140, opened 2017-12-29
OS X Command+V


neseleznev:
_geocode is not invoking while user pastes searchInput via Cmd+V on OS X.

The issue is supposed to by fixed in the following line
https://github.com/mapbox/mapbox-gl-geocoder/blob/53d06d871574a227a6b414f3370247ea16e29238/lib/index.js#L119
since Cmd is a metaKey.

Probably, it could be checked by something like e.code === 'MetaLeft' || e.code === 'MetaRight' ||  e.code === 'OSLeft' || e.code === 'OSRight';

I'm not a web developer, correct me if I'm wrong

andrewharvey:
Which version of mapbox-gl-geocoder are you using?

Does it not open the suggestions list after pasting or just it won't paste into the input at all?

The latest version fixed paste events with https://github.com/tristen/suggestions/pull/17 see https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CHANGELOG.md could you confirm you're still seeing this with v2.1.2?

ingalls:
Stale - Feel free to reopen if you are still having problems.

matrad:
Still seeing this issue on v2.2.0 with Chrome 67.0.3396.99 x64 (newest build) and OSX 10.13.3

gtk-hs:
Hey all! I'm currently seeing this issue on modern FF(68.0 )/chrome(75.0.3770.142)/Safari(12.1.2) OS X on our production site, and even on the mapbox website: 

https://docs.mapbox.com/help/tutorials/local-search-geocoding-api/

I've tried manually dispatching the change/keyup/keydown event on the target input field, but can't get it to populate the geocode dropdown UI.

This can be a painful UX experience for our users that are confused as to why pasting a valid address returns no response in the UI.

If you can get the dropdown to trigger on paste, that would be ideal, as users will see immediate feedback and pick the proper result. Also if hitting the enter key while focused on that dropdown kicked off the geocoder dropdown UI I think that would help as well (although hitting any OTHER key does appear to have the desired effect). While I'm at it I feel clicking the "search" svg icon when text is pasted in should probably kick off that UI as well (although it might be unnecessary if the paste event worked in the first place).

In the meanwhile, is there anyway I can programmatically kick off the dropdown of geocode results so I can reduce friction on end users who are having a hard time figuring this out?

Grsmto:
[...]

PR #158, opened 2018-07-11
(fix) COMMAND V on OSX

matrad:
issue #140

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -129,8 +129,12 @@ MapboxGeocoder.prototype = {
     }
 
     // TAB, ESC, LEFT, RIGHT, ENTER, UP, DOWN
-    if (e.metaKey || [9, 27, 37, 39, 13, 38, 40].indexOf(e.keyCode) !== -1) return;
-
+    if(e.metaKey && e.code != 'KeyV') {
+      return;
+    } else if ([9, 27, 37, 39, 13, 38, 40].indexOf(e.keyCode) !== -1) {
+      return;
+    }
+    
     if (target.value.length >= this.options.minLength) {
       this._geocode(target.value);
     }


andrewharvey:
Since I don't have OSX, I can't test it, so I'll leave it for someone at Mapbox to look at but I wanted to mention that according to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

> The KeyboardEvent.code property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value which isn't altered by keyboard layout or the state of the modifier keys.

KeyV is only the character V on the QWERTY layout, so I'm not sure it can be relied on here... Have you tested what happens on other keyboard layouts?

matrad:
@andrewharvey the osx devices mostly have the Apple-Keyboard, so for the OSX devices this will work.
On other devices I've not tested it.

But here may be other investigation needed, because this meta key can also be Windows+V on Windows-Devices:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey

On OSX Devices I've tested it & it's working this way.

andrewharvey:
Yeah but you can change your keyboard layout on OSX even with the Apple-keyboard. So on OSX using Dvorak the paste shortcut is physically Cmd+. because that's where V is.

yapalexei:
Out of curiosity would the `_onChange` method be a better candidate for resolving this issue? because the change can be throttled/debounced just as well and accomplish a broader set of use cases.

Grsmto:
I'm confused as on why the `paste` event is not used for this instead of hardcoded keycodes?

andrewharvey:
The paste event is used in the upstream suggestions library.

andrewharvey:
https://github.com/tristen/suggestions/blob/9328f1f3d21598c40014892e3e0329027dd2b538/src/suggestions.js#L41-L43

On Linux Chromium, right click paste, Ctrl+V and middle click paste all work and open the suggestions box. Happy to see a fix for OSX, but as mentioned above from KeyV isn't acceptable since on different keyboard layouts it doesn't represent paste.

Grsmto:
Ok I took a deeper look at this and I think I get it now, thanks for your clarifications @andrewharvey 

The problem is that listening to keycodes on `keyDown` event is not a sustainable solution as it's based on device/os/browsers. My assumption is that the author didn't use the `onChange` event is because [the upstream `Suggestions` library is triggering](https://github.com/tristen/suggestions/blob/9328f1f3d21598c40014892e3e0329027dd2b538/src/suggestions.js#L178) a fake `onChange` event when user selects an address. Which would then trigger a geocode call. So author thought of using the `keyDown` event instead.

My suggestion would be to not use the `keyDown` event at all and instead either:
- use the [input event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) instead
- use the `change` event but check for `Event.isTrusted` so then we can filter out the event triggered by the `Suggestions` library

You opened this a year ago so I suppose you're not working on this anymore. Let me know if you want me to look into these solutions.

andrewharvey:
>The problem is that listening to keycodes on keyDown event is not a sustainable solution as it's based on device/os/browsers. My assumption is that the author didn't use the onChange event is because the upstream Suggestions library is triggering a fake onChange event when user selects an address. Which would then trigger a geocode call. So author thought of using the keyDown event instead.
>
>My suggestion would be to not use the keyDown event at all and instead either:
>
>use the input event instead
>use the change event but check for Event.isTrusted so then we can filter out the event triggered by the Suggestions library

Great analysis, thank you!

>You opened this a year ago so I suppose you're not working on this anymore. Let me know if you want me to look into these solutions.

I think that's a fair assumption, I'd be keen to help review a PR with a fix if you're interested in working on it.

andrewharvey:
closing in favour of #140 since it's not dependent on a single keyboard layout.

PR #163, opened 2018-08-21
Amend the client initialization to pass through options

joshwarwick:
When trying to use a custom endpoint (for example tunnelling a request through some backend) the client gets initialized without passing the `endpoint` parameter to `mapbox-gl`. This fixes this problem, also wondered if there is any gotchas I'm missing here? 

Thanks!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -57,7 +57,7 @@ MapboxGeocoder.prototype = {
 
   onAdd: function(map) {
     this._map = map;
-    this.mapboxClient = new MapboxClient(this.options.accessToken);
+    this.mapboxClient = new MapboxClient(this.options.accessToken, this.options);
     this._onChange = this._onChange.bind(this);
     this._onKeyDown = this._onKeyDown.bind(this);
     this._onQueryResult = this._onQueryResult.bind(this);
--- a/lib/index.js
+++ b/lib/index.js
@@ -57,7 +57,10 @@ MapboxGeocoder.prototype = {
 
   onAdd: function(map) {
     this._map = map;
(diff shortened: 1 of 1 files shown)


andrewharvey:
Thanks, I agree we need to expose the origin/endpoint option.

I'm thinking it's better to explicitly add origin/endpoint as an option instead of passing the whole options.

The side affect of that means when we swap over to @mapbox/mapbox-sdk it isn't a breaking change (`endpoint` vs `origin`).

https://github.com/mapbox/mapbox-sdk-js/blob/v1.0.0-beta10/API.md#mapboxclient
https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/classes.md#mapiclient

In line with @mapbox/mapbox-sdk we should call this `origin`.

joshwarwick:
Thanks for the speedy response! 

Yeah did originally think this before following line 173 which passed the options directly to the `geocodeForward` function. Not sold on the above commit to pass through options that are undefined but makes sense going forward as to not introduce a breaking change!

andrewharvey:
Right now the MapboxClient option is `endpoint` (https://github.com/mapbox/mapbox-sdk-js/blob/v1.0.0-beta10/API.md#mapboxclient) but in the current Mapbox SDK it's called `origin` (https://github.com/mapbox/mapbox-sdk-js/blob/master/docs/classes.md#mapiclient), so let's call this `origin`.

Could you add documentation for the option (see https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L10) and then run `npm run docs` to update API.md?

PR #166, opened 2018-09-24
Fix API markdown to format event list properly

johallar:
`clear` event is not displaying on a new line, which makes it very easy to miss. Adding newline so the `clear` event shows up in the list properly.

### DIFF
--- a/API.md
+++ b/API.md
@@ -16,7 +16,7 @@ A geocoder component using Mapbox Geocoding API
 
 **Parameters**
 
--   `options` **[Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object)** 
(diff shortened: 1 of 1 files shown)


andrewharvey:
See the background at https://github.com/mapbox/mapbox-gl-geocoder/pull/125, it's a bug in the upstream documentation generator. The API.md is automatically generated so this means we either:

1. Fix upstream
2. Manually fix it every time the API.md is generated

I don't like the idea of manually fixing this everytime, but given we've had multiple PRs it does seem like an issue people notice, so I guess we can try to fix it manually.

What do you think?

johallar:
Agree it would be best to fix upstream and I haven't been able to find an easy way to modify the comment to get it displaying properly in the short term.

I'll leave it up to you to merge this or not, and will continue looking into either an upstream fix or a syntax change in the comments to get it rendering properly.

PR #173, opened 2019-01-14
Add travis + update tests

### LINKED ISSUE #109: Drop circleci
It doesn't support WebGL.

rmrice:
This PR adds a travis.yml to run smokestack headlessly via `xvfb` and run tests. I also disabled CircleCI.

Additionally, I updated two tests because they were failing. One was because a the top result for the query changed (`London Market` vs `London Street`), the second was due to the country bbox exceptions test failing. Canada is the only country listed as an exception in `./exceptions.js` that returns a truthy result for that test. I am not super familiar with the history of Mapbox's geocoding API, but maybe the bbox results have changed since these tests were last updated. 

Closes #109


andrewharvey:
Looks good to me @rmrice, it's good to see CI actually running the unit tests.

PR #175, opened 2019-01-16
Port to mapbox-sdk, support reverse geocoding, custom endpoint

### LINKED ISSUE #73: Use mapbox-sdk-js
Currently this module does its own URL construction and supports a limited set of params. Rely on mapbox-sdk-js instead.

### LINKED ISSUE #156: upgrade to new @mapbox/mapbox-sdk
https://github.com/mapbox/mapbox-sdk-js

Currently using the old unmaintained version, we should upgrade to the latest.

### LINKED ISSUE #168: Reverse Geocoding
Reverse geocoding does not work unless the limit is set to 1.
I get a http status 422:
https://api.mapbox.com/geocoding/v5/mapbox.places/-122.1162864%2C37.4133183.json?access_token=xxx&limit=5

However if I set the limit to 1 it works but then my forward geocoding is also limited to 1 which is unwanted.

`var geocoder = new MapboxGeocoder({
accessToken: mapboxgl.accessToken,
limit:1
});
geocoder.query("-122.1162864,37.4133183");`

rmrice:
This PR

- Replaces old `mapbox` as client with new [mapbox-sdk](https://github.com/mapbox/mapbox-sdk-js) (closes #156)
- Supports reverse geocoding (closes #168 & #121 )
- Supports custom endpoint by passing it in as `origin` in `options` (closes https://github.com/mapbox/mapbox-gl-geocoder/pull/163)
- Formats code for readability (I used [prettier](https://prettier.io/))
- Updates API docs

Apologies for the big diff -- I would have split some of this up into two PRs but I had to fix reverse geocoding in order to make tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
-  - '4'
   - '6'
-  - '8'
 env:
   global: MapboxAccessToken
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: node_js
-node_js: '4'
+node_js:
+  - '4'
+  - '6'
+  - '8'
 env:
   global: MapboxAccessToken
 addons:
--- a/API.md
+++ b/API.md
@@ -40,7 +40,7 @@ A geocoder component using Mapbox Geocoding API
     -   `options.language` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Specify the language to use for response text and query result weighting. Options are IETF language tags comprised of a mandatory ISO 639-1 language code and optionally one or more IETF subtags for country or script. More than one value can also be specified, separated by commas.
     -   `options.filter` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function which accepts a Feature in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format to filter out results from the Geocoding API response before they are included in the suggestions list. Return `true` to keep the item, `false` otherwise.
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
--   `config`  
+    -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
 
 **Examples**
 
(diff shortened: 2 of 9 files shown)


andrewharvey:
Nice job tackling this @rmrice! I do want to take a look, but I won't get a chance until next week, if you are happy to wait :wink: .

rmrice:
Thanks y'all! 🙏

andrewharvey:
@rmrice Should we make this new reverse geocoding feature optional?

I was -1 to implementing it which I voiced in #168, but since it's now implemented could we have an option to disable it, or disable it by default and have an option to enable it (to ensure backwards compatibility)?

As a user for most apps I'd expect entering a lat/lon would simply forward geocode that lat/lon to the exact location, rather than listing nearby features from the reverse geocoder.

rmrice:
@andrewharvey we will need to make reverse geocoding available by default if you want to use coordinates as a query. Reverse geocoding (querying with coordinates) didn't work at all with the plugin previously. Any time you entered coordinates you would receive an error from the geocoding API since the request wasn't handled properly (this may not have been true with a previous version of our geocoding API, but with the current version it doesn't work). Implementing reverse geocoding doesn't introduce a breaking change, so I don't think we need to worry about backwards compatibility. 

> As a user for most apps I'd expect entering a lat/lon would simply forward geocode that lat/lon to the exact location, rather than listing nearby features from the reverse geocoder.

What you're describing is reverse geocoding. The reason the API limits results to 1 (the error discussed in #168) is because it expects to return an exact location, not nearby locations. Right now it returns the placename of the coordinates as the main result (assuming you entered them as `lat, lon`), but also gives you query options in case you entered lon/lat:

screenshot 2019-01-21 15 08 26

I definitely think we can improve the results (like assume lat/lon, or make this configurable within the plugin options, or put the main result on top), but at least this enables you to search for coordinates. Happy to help improve this.

andrewharvey:
> @andrewharvey we will need to make reverse geocoding available by default if you want to use coordinates as a query. Reverse geocoding (querying with coordinates) didn't work at all with the plugin previously.

Isn't that what https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ does? That example is still using 2.3.0 and it's using coordinates as a query, and works fine for me to search coordinates. ie. it was working back in 2.3.0, just needed some configuration on the user side.

In that example with 2.3.0 if I enter "150.854, -34.667" I get taken to Kiama exactly to the coordinates specified.

Same example with 3.0.1 if I enter "150.854, -34.667" I now get two results, the coordinates and Kanangra. But Kanangra is 100km away. ...Hang on, actually it looks like this happens due to  a bug. It's using `parseInt` which is rounding the coordinate to an integer and loosing all the precision.

selection_999 183

At the moment with 3.0.1 I can't disable listing "Kanangra" as an option to get the same results as 2.3.0.

Even without that issue though, the reverse geocoder will still provide results a long way away. For example neither of these reverse geocoder results are very close to the coordinates entered.

selection_999 184

rmrice:
@andrewharvey I see what you mean. I can try to get to fixing this this week. In terms of suggestions:

- use `parseFloat` rather than `parseInt` (oops)
- will make `reverseGeocode` an option rather than assumed if entering coordinates

Let me know if anything else pops up!

PR #177, opened 2019-01-23
Make reverse geocoding optional

### LINKED ISSUE #120: allow spaces in comma separated types option
mapboxgl - 0.37.0
geocoder - 2.1.0

The request documentation states that you can pass multiple types separated by a comma into the options: [API.md](https://www.mapbox.com/api-documentation/#request-format) but the following error is thrown if more than one is passed in.

`Uncaught (in promise) TypeError: Cannot read property 'length' of undefined
    at r.<anonymous> (mapbox-gl-geocoder.min.js:1)
    at <anonymous>`

I have found that this is caused if a space is present between the types:

`var m = new MapboxGeocoder({
        types: "district, place"
})`

It would be great when parsing the types if each value was trimmed and/or the documentation updated to show the proper formatting.

Thanks.

rmrice:
Based on concerns addressed in https://github.com/mapbox/mapbox-gl-geocoder/pull/175#issuecomment-456200161, this PR should help with backwards compatibility for some of our public examples that use the plugin.

- Makes `reverseGeocode` an option (defaults to true)
- Fixes string parsing while passing a list in to options when the client expects an array (for `countries`, `types`, and `languages`)
- Adds a test for 👆 to make sure that this is being done correctly
- Fixes coordinate parsing (if `reverseGeocode` is enabled)
- Appends `config` to `response` (so the config passed to the client is testable)

@andrewharvey now to use with https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/, you could set `reverseGeocode: false` while initializing the geocoder. Doing so gives you the same result as the example using 2.3.0. However, it will still throw an error in the browser console since any coordinate-like query you send to our geocoding API attempts to reverse geocode (which is expected). I'm not sure the best way to handle this, but I think it's fine for now since the local geocoder still returns the same result and the console error doesn't affect the result. Let me know if you have any ideas.

closes #120

### DIFF
--- a/API.md
+++ b/API.md
@@ -35,14 +35,13 @@ A geocoder component using Mapbox Geocoding API
     -   `options.types` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** a comma seperated list of types that filter
         results to match those specified. See <https://www.mapbox.com/developers/api/geocoding/#filter-type>
         for available types.
-        If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used. If no type is specified, reverse geocodes will default to 'district'.
-    -   `options.minLength` **[Number](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number)** Minimum number of characters to enter before results are shown. (optional, default `2`)
+        If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used.
     -   `options.limit` **[Number](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number)** Maximum number of results to show. (optional, default `5`)
     -   `options.language` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Specify the language to use for response text and query result weighting. Options are IETF language tags comprised of a mandatory ISO 639-1 language code and optionally one or more IETF subtags for country or script. More than one value can also be specified, separated by commas.
     -   `options.filter` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function which accepts a Feature in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format to filter out results from the Geocoding API response before they are included in the suggestions list. Return `true` to keep the item, `false` otherwise.
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
     -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
-    -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to true. Expects coordinates to be lon, lat.
+    -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to false. Expects coordinates to be lat, lon.
 
 **Examples**
 
--- a/API.md
+++ b/API.md
@@ -35,12 +35,14 @@ A geocoder component using Mapbox Geocoding API
     -   `options.types` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** a comma seperated list of types that filter
         results to match those specified. See <https://www.mapbox.com/developers/api/geocoding/#filter-type>
         for available types.
+        If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used. If no type is specified, reverse geocodes will default to 'district'.
     -   `options.minLength` **[Number](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number)** Minimum number of characters to enter before results are shown. (optional, default `2`)
     -   `options.limit` **[Number](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Number)** Maximum number of results to show. (optional, default `5`)
     -   `options.language` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)?** Specify the language to use for response text and query result weighting. Options are IETF language tags comprised of a mandatory ISO 639-1 language code and optionally one or more IETF subtags for country or script. More than one value can also be specified, separated by commas.
     -   `options.filter` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function which accepts a Feature in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format to filter out results from the Geocoding API response before they are included in the suggestions list. Return `true` to keep the item, `false` otherwise.
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
     -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
+    -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to true.
 
 **Examples**
 
--- a/API.md
+++ b/API.md
@@ -42,7 +42,7 @@ A geocoder component using Mapbox Geocoding API
     -   `options.filter` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function which accepts a Feature in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format to filter out results from the Geocoding API response before they are included in the suggestions list. Return `true` to keep the item, `false` otherwise.
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
     -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
-    -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to true.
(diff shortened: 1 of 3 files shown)


andrewharvey:
Cool, thanks @rmrice. I'll add some comments.

It's easier to review if unrelated changes are done as different PRs instead of one big one, but I know it can be eaiser sometimes to just bundle everything together.

lib/index.js
```diff
@@ -185,33 +188,43 @@ MapboxGeocoder.prototype = {
...
     // Create config object
     var config = keys.reduce(function(config, key) {
       if (self.options[key]) {
         // countries, types, and language need to be passed in as arrays to client
         // https://github.com/mapbox/mapbox-sdk-js/blob/master/services/geocoding.js#L38-L47
-        ['countries', 'types', 'language'].indexOf(key) > -1 ?
-          config[key] = [self.options[key]]
-        : config[key] = self.options[key]
+        ['countries', 'types', 'language'].indexOf(key) > -1
+          ? (config[key] = self.options[key].split(', '))
```
andrewharvey: Would this close #120 by allowing spaces?

In mapbox-gl-geocoder v2.3.0 you could pass a comma separated string without spaces, for backwards compatibility I think we should continue to support this.

test/test.geocoder.js
```diff
@@ -109,6 +109,77 @@ test('geocoder', function(tt) {
...
+      })
+    );
+  });
+
+  tt.test('parses options correctly', function(t) {
+    t.plan(4);
+    setup({
+      language: 'en, es, zh',
+      types: 'district, locality, neighborhood, postcode',
+      countries: 'us, mx'
```
andrewharvey: we should add a unit test for non spaces. eg. `'district,locality'` to ensure this is still supported and backwards compatible

lib/index.js
```diff
@@ -185,33 +188,43 @@ MapboxGeocoder.prototype = {
...
     var request;
     // check if searchInput resembles coordinates, and if it does,
     // make the request a reverseGeocode
-    if (/^(\-?\d+(\.\d+)?),\s*(\-?\d+(\.\d+)?)$/.test(searchInput)) {
+    if (
+      this.options.reverseGeocode &&
+      /(-?\d+\.?\d*)[, ]+(-?\d+\.?\d*)[ ]*$/.test(searchInput)
+    ) {
       // parse coordinates
       var coords = searchInput.split(',').map(function(c) {
```
andrewharvey: This breaks down if I query with a space since the regex accepts it but this splitting code doesn't.

eg. query `10 10`

andrewharvey:
> ...However, it will still throw an error in the browser console since any coordinate-like query you send to our geocoding API attempts to reverse geocode (which is expected). I'm not sure the best way to handle this, but I think it's fine for now since the local geocoder still returns the same result and the console error doesn't affect the result. Let me know if you have any ideas.

That's true, I agree it's probably fine for now. In the future on the client side we could detect an input of `number,number` and if reverseGeocode is false then don't make the remote request at all.

API.md
```diff
@@ -35,12 +35,14 @@ A geocoder component using Mapbox Geocoding API
...
     -   `options.filter` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function which accepts a Feature in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format to filter out results from the Geocoding API response before they are included in the suggestions list. Return `true` to keep the item, `false` otherwise.
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
     -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
+    -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to true. Expects coordinates to be lon, lat.
```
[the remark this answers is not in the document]
rmrice: I think if the documentation specifies `lon, lat` then this is fine. That is what is passed into the geocoder and what one would expect when thinking about coordinates in terms of `x,y`.

We have users who expect reverse geocoding to work out of the box with this plugin, so the intent here is to simply enable it. The use case is the same as what https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/  aims to achieve-- enter a coordinate, and fly to location of coordinates entered. Do you think it would be better to disable it by default, so the user can be intentional about enabling it and which order the enter coordinates in? That way we don't need to modify the local geocoder example.

lib/index.js
rmrice: Yes, would close #120. Fixed in latest commit.

test/test.geocoder.js
rmrice: Done 👍

lib/index.js
rmrice: Fixed in latest commit

API.md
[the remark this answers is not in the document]
andrewharvey: I think it would be a good change for this plugin to support searches of lat/lon coordinates directly, that was certainly what the example https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ was designed for. Allowing the user to enter a WGS84 coordinate and have the map go to that exact coordinate, seems like a sensible default to me for most maps.

There would be cases when this is not wanted, so could optionally be disabled. eg. Using mapbox-gl-geocoder to just search postcodes only. In this way it kind of becomes an extra type (like country,postcode,locality,address...), but since the "coordinate" type nor forward coordinate searches aren't supported by the Mapbox Geocoding API, we need to implement it client side.

The Mapbox Geocoding reverse geocodes API lets you identify the geographic places at/around a coordinate, but doesn't give back that exact coordinate. [example](https://docs.mapbox.com/search-playground/#{%22url%22:%22%22,%22index%22:%22mapbox.places%22,%22staging%22:false,%22onCountry%22:true,%22onType%22:true,%22onProximity%22:true,%22onBBOX%22:true,%22onLimit%22:true,%22onLanguage%22:true,%22countries%22:[],%22proximity%22:%22%22,%22typeToggle%22:{%22country%22:false,%22region%22:false,%22district%22:false,%22postcode%22:false,%22locality%22:false,%22place%22:false,%22neighborhood%22:false,%22address%22:false,%22poi%22:false},%22types%22:[],%22bbox%22:%22%22,%22limit%22:%22%22,%22autocomplete%22:true,%22languages%22:[],%22languageStrict%22:false,%22onDebug%22:false,%22selectedLayer%22:%22%22,%22debugClick%22:{},%22query%22:%22150,-35%22}).

Which for a search box, as a user I would expect that entering coordinates would take me to that exact coordinate, which is what the example at https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ does, but not what this new reverseGeocode mode does.

I guess that's just because I'm a more technical user, for a casual user who doesn't know anything of coordinates, getting back a place name in the search list is better than getting back coordinates. But still, if they enter exact coordinates, I would expect the map to go to those exact coordinates. Hence how I came to the conclusion that I would usually turn reverseGeocode off.

screenshot from 2019-01-24 22-50-52

Back to the point on coordinate pair order, I would say that in general mapping/gps it's predominantly lat, lon order. I've only ever seen WGS84 in lon, lat order in the GeoJSON spec and all other things which stem from that.

In fact Google Maps only supports lat,lon order not lon,lat. I still think for most general purpose apps it's the wrong chose to only accept lon,lat, something like https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ which supports both is better in my view, but more complicated for non-technical users.

That's just my thoughts, I don't know what's best, I can only try to provide some feedback on how I'd expect users to expect mapbox-gl-geocoder to work.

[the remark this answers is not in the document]
briandaviddavidson: > I am keen to try to understand the use case for this kind of reverse geocode feature, would it be if you want to find nearby poi's by searching coordinates, or just giving some context to the coordinate you entered (eg. country, state, city, locality, address)?

Most of the time, this comes when customers want to have their users drop a point, then get POIs or other features around them. For scenarios like that, we recommend reverse geocoding. For scenarios like that, our users expect that if a point is dropped, they will get some sort of context back. I have seen that use case, but not very frequently. I think most users begin by typing something in like a place, but it mostly depends on use case.

> Do you think it would be better to disable it by default, so the user can be intentional about enabling it and which order the enter coordinates in? That way we don't need to modify the local geocoder example.

I think this makes sense. Disabled by default, then purposfully making the user select `long/lat` or `lat/long`. Then it's a consious decision on the implementer's part as to how thier users expect to put in coordinates.

> Which for a search box, as a user I would expect that entering coordinates would take me to that exact coordinate, which is what the example at https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ does, but not what this new reverseGeocode mode does.

Hmm, I see the point here. I can imagine a more technical user wouldn't necessarily want a POI, but the exact coordinates. I wonder if there is a way to implement that as a feature flag in this as well. So you could enable `reverseGeocoding`, then tell it what order, and if coordinates should appear in the drop down. After writing it out just now, it sounds like it might be better suited for another PR, but that could be a nice addition to the `reverseGeocoding`.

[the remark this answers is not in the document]
rmrice: > Which for a search box, as a user I would expect that entering coordinates would take me to that exact coordinate, which is what the example at https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ does, but not what this new reverseGeocode mode does.

I am pretty sure that the scope of the reverseGeocoding result is based on `types` specified by the config, which is one of `'country'|'region'|'postcode'|'district'|'place'|'locality'|'neighborhood'|'address'|'poi'|'poi.landmark'`, and it returns one result based on the type set.

I am going to have it default to `poi.landmark` (most specific type), but I've noticed if there's not a specific poi or landmark associated with the coordinate, it will return a feature that is most closely associated with that point. I think that's just how our geocoder works 🤷‍♀️ 

I just pushed some more commits, which I'll outline in another comment below.

rmrice:
I just pushed some stuff up based on suggestions:

- `reverseGeocode` is false by default
- `reverseGeocode` accepts coordinates as `lat, lon`, and is forgiving if not separated by a comma
- for reverse geocodes, if `types` is specified in the config, the first one will be used (required by the API), if no type is specified, it will default to `poi.landmark`
- fixed bugginess interpolating options required as an array (countries, types, languages)
- updated docs

would like to get this in & publish a new release in order to fix the bugs #175 introduced 🙏 any other systemic opinions on how the geocoder plugin should function I think need to be more widely discussed.

lib/index.js
```diff
@@ -219,7 +218,11 @@ MapboxGeocoder.prototype = {
       // parse coordinates
       var coords = searchInput.split(/[\s(,)?]+/).map(function(c) {
         return parseFloat(c, 10);
-      });
+      }).reverse();
+
+      // client only accepts one type for reverseGeocode, so
+      // use first config type if one, if not default to poi.landmark
+      config.types ? [config.types[0]] : ["poi.landmark"];
```
andrewharvey: nit picking here but according to https://docs.mapbox.com/api/search/#reverse-geocoding `poi.landmark` is deprecated, returning the same as `poi`, so probably better to just change this to `poi`.

andrewharvey:
@rmrice Looks good to me. Cheers.

API.md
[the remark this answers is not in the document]
andrewharvey: > Most of the time, this comes when customers want to have their users drop a point, then get POIs or other features around them. For scenarios like that, we recommend reverse geocoding and our users expect that if a point is dropped, they will get some sort of context back. I have seen that use case, however, I think most users begin by typing something in like a place. It definitely depends on what the end user is trying to accomplish.

I guess I would implement that separately, pass it a coordinate programatically and it populates a list of nearby POI's which link to the map.

Either way, I don't want to come across as trying to dictate the direction of this plugin, just chiming in my thoughts based on how I've seen it used, and how I've used it.

> Hmm, I see the point here. I can imagine a more technical user wouldn't necessarily want a POI, but the exact coordinates.

Exactly. I guess if you're entering coordinates in the first place, you at least a somewhat technical user?

> I wonder if there is a way to implement that as a feature flag in this as well. So you could enable reverseGeocoding, then tell it what order, and if coordinates should appear in the drop-down as the first result. After writing it out just now though, it sounds like it might be better suited for another PR, but that could be a nice addition to the reverseGeocoding.

So like https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-local-geocoder/ , but out of the box?

PR #180, opened 2019-01-31
ensure proximity option is passed through to request

andrewharvey:
this fixes a regression in v3 which broke the proximity option, including a test


andrewharvey:
@rmrice Should we do a patch release to get this out?

rmrice:
@andrewharvey yes, let's do that. I can roll it out.

Issue #99, opened 2017-02-15
result event is invoked twice


andrewharvey:
http://jsbin.com/nalopux/edit?html,console,output

If I type something in the Geocoder then select an option, the `result` event is seemingly invoked twice with the same result.

andrewharvey:
This is an upstream issue in suggestions

brambow:
This is causing issues for me in a situation where I am trying to animate some layers that contain the resulting point of a geocode operation.  Since the result is invoked twice, my animation runs twice and looks weird.  Any ideas for a workaround appreciated.

andrewharvey:
@brambow Do you mind testing the proposed solution at https://github.com/tristen/suggestions/issues/13

> If I disable the dispatch of the change event at https://github.com/tristen/suggestions/blob/gh-pages/src/suggestions.js#L151 the event is only triggered once and there doesn't seem to be any side affects.

I didn't notice any issue, if you don't either we can try to get a fix for this!

brambow:
Yep.  That seems to fix it.

ckc11:
This also fixed it for me as well.

andrewharvey:
Just noting that removing this code does fix it for modern browsers but breaks it for IE11. https://github.com/tristen/suggestions/issues/13#issuecomment-313595481

jwhazel:
Throwing in my request for a fix as well (preferably one that satisfies IE11). My current solution is to set up global to store the `latlng` and check to see if the new event is different than the last. Clearly not the most elegant solution.

```lastGeocode = "";
geocoder.on('result', function (e) {
	if(e.result.center.toString() !== lastGeocode){
		drawCustomMarker(e.result.center, e.result.place_name);
	}

	lastGeocode = e.result.center.toString();
});```

pathmapper:
Using GL JS `v0.48.0` and Geocoder  `v2.3.0`, it looks like the behavior changed a bit compared to https://github.com/mapbox/mapbox-gl-geocoder/issues/99#issue-207780402:

The second `result` event is now fired after a click on the map (Chrome on Linux): 
https://jsbin.com/dapurelime/edit?html,console,output

andrewharvey:
> The second result event is now fired after a click on the map (Chrome on Linux):

That's interesting, I'm also running Chrome on Linux and can't replicate that, in fact I'm only ever seeing the result event emitted once. :thinking:

pathmapper:
[...]

PR #183, opened 2019-02-15
Interaction event logging

scottsfarley93:
This PR adds end-user interaction logging via the mapbox events api. Specifically, it instruments the plugin with two events: search-select and search-start.

- `search-select`: this event is triggered when an item from the dropdown is selected by the user.  The item selected by the end user, the input text string and any options used to make the search request are logged.
- `search-start`: this event is triggered any time the user starts with an empty search bar and starts a search session. Input text and any options used to make the search are logged. 

### Why log interaction events?
Anonymized logging of search interaction events will give the mapbox search team a better understanding of how end users are using the search service. Specifically, instrumenting the plugin with selection and start events will help us improve our relevance algorithm and reduce failed searches. In general, these events will help us understand usage patterns and tune the search engine for more relevant results in the future. 

For each event, the following attributes are recorded:

- timestamp: the current time
- session identifier:   a randomly generated string that is generated upon plugin instantiation
- country: the `options.country` value set on the geocoder
- userAgent: the browser’s user agent string
- language: the `options.language` value set on the geocoder
- bbox: the `options.bbox` value set on the geocoder
- types: the `options.types` value set on the geocoder
- proximity: the `options.proximity` value set on the geocoder
- limit: the `options.limit` value set on the geocoder
- queryString: the text the end user typed into the search box
- mapCenterLat: the latitude of the center of the current map view
- mapCenterLng: the longitude of the center of the current map view
- mapZoom: the zoom of the current map view
- keyboardLocale: the default language of the user’s browser
- (for search.select events) the result index, as numeric array index, will also be reported

End user geo location is not recorded. Session IDs are recorded, but they are randomly generated at the instantiation of a new gl-geocoder object and cannot be related to a particular user.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,0 +1,184 @@
+'use strict';
+
+var request = require('request');
+var crypto = require('crypto');
+
+/**
+ * varruct a new mapbox event client to send interaction events to the mapbox event service
+ * @param {Object} options options with which to create the service
+ * @param {String} options.accessToken the mapbox access token to make requests
+ */
+function MapboxEventManager(options){
+  this.endpoint = 'https://api.mapbox.com/events/v2';
+  this.access_token = options.accessToken;
+  this.sessionID = this.generateSessionID();
+  this.userAgent = this.getUserAgent();
+  this.version = '0.0.1'
+  // parse global options to be sent with each request
+  this.countries =  (options.countries) ? options.countries.split(",") : null;
+  this.types =  (options.types) ? options.types.split(",") : null;
+  this.bbox = (options.bbox) ? options.bbox : null;
+  this.language = (options.language) ? options.language.split(",") : null;
+  this.limit = (options.limit) ? +options.limit : null;
+  this.locale = navigator.language || null;
+  // keep some state to deduplicate requests if necessary
+  this.lastSentInput = "";
+  this.lastSentIndex = 0;
+}
+
+MapboxEventManager.prototype = {
+
+  /**
+     * Send a search.select event to the mapbox events service
+     * This event marks the array index of the item selected by the user out of the array of possible options
+     * @param {Object} selected the geojson feature selected by the user
+     * @param {Object} geocoder a mapbox-gl-geocoder instance
+     * @returns {Promise}
+     */
+  select: function(selected, geocoder, callback){
+    var resultIndex = this.getSelectedIndex(selected, geocoder);
+    var payload = this.getEventPayload('search.select', geocoder);
+    payload.resultIndex = resultIndex;
+    if (resultIndex === this.lastSentIndex && payload.queryString === this.lastSentInput) {
+      // don't log duplicate events if the user re-selected the same feature on the same search
+      if (callback) return callback();
+    }
+    this.lastSentIndex = resultIndex;
+    this.lastSentInput = payload.queryString;
+    return this.send(payload, callback)
+  },
+
+  /**
+     * Send a search-start event to the mapbox events service
+     * This turnstile event marks when a user starts a new search
+     * @param {Object} geocoder a mapbox-gl-geocoder instance
+     * @returns {Promise}
+     */
+  start: function(geocoder, callback){
+    var payload = this.getEventPayload('search.start', geocoder);
+    return this.send(payload, callback);
+  },
+
+  /**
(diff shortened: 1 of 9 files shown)


lib/events.js
```diff
@@ -0,0 +1,184 @@
...
+    var sha = crypto.createHash('sha256');
+    sha.update(Math.random().toString());
+    return sha.digest('hex')
+  },
+
+  /**
+     * Get a user agent string to send with the request to the events service
+     */
+  getUserAgent: function(){
+    return 'mapbox-gl-geocoder.' + this.version + "." + navigator.userAgent;
```
andrewharvey: When I test this locally with `yarn run start` it reports "undefined" here. Not sure if that's intended or not?

andrewharvey:
hi @scottsfarley93 I made some comments/feedback for your consideration.

mapbox-gl-geocoder has a longstanding bug #99 that causes the 'results' event to be emitted twice when selecting a result. That's also causing duplicate `search-select` events to be logged. So unless the bug is fixed, something to keep in mind when looking at results on the receiving end :wink: 

---

> * `search-start`: this event is triggered any time the user starts with an empty search bar and starts a search session. Input text and any options used to make the search are logged.

If the user continues to refine their search, nothing is subsequently logged. Was that the intention?

eg. Click the search box and start typing "white", typed quickly enough this input "white" is logged as search-start. Pause then continue to refine the search with "whitehouse" pause, nothing else is logged.

---

Do you plan on treating results from the `localGeocoder` as a special case?

At the moment if someone uses `localGeocoder` you're still logging their `search-select` but the results aren't going to mean anything, since the resultIndex is logged, but you have no visibility on which non-Mapbox search results were present. If not accounted for, you could incorrectly think someone is selecting results number 4 wondering why your geocoder isn't giving them expected result as one, but actually localGeocoder results are 1-5 so not even showing any Mapbox results anyway.

The same logic applies when `filter` is in use.

One option is to to skip logging when these flags are enabled, I'm not sure how you'd be able to gather any insights on the server side when these flags are present...?

---

I think event logging should only be enabled when the origin is set to Mapbox.

scottsfarley93:
hey @andrewharvey -- thank you for the thoughtful comments. I have updated the PR accordingly. 

1. I noticed this -- I am deduplicating prior to sending the event by checking the last request made [here](https://github.com/mapbox/mapbox-gl-geocoder/pull/183/files#diff-71dcd327d0ca067b490b22d677f81966R44). This just checks that the next sent event is not the same input string and result index as the previous event. 
2. Yes this is the intended behavior. The `search.start` event is designed to give us some insight into how often new searches are started from an empty search bar. In the future, we may wish to log subsequent events (e.g., after additional query refinement), but for now we're happy with just knowing that a search has been initiated with the given arguments.
3. Excellent point about the `filter` and `localGeocoder` options. I have disabled logging for gl-geocoder instances that use these options [ref](https://github.com/mapbox/mapbox-gl-geocoder/pull/183/files#diff-71dcd327d0ca067b490b22d677f81966R192). 
4. As with above, I've disabled logging for instances that use origins other than api.mapbox.com [ref](https://github.com/mapbox/mapbox-gl-geocoder/pull/183/files#diff-71dcd327d0ca067b490b22d677f81966R193). 
5. I fixed the userAgent string that you noted to be `undefined`, so that it's not undefined any more and added a test for it.

pathmapper:
Hi @scottsfarley93 ,

> End user geo location is not recorded.

I think it is recorded if there is a [GeolocateControl](https://docs.mapbox.com/mapbox-gl-js/api/#geolocatecontrol) in active state:

> active - the map's camera automatically updates as the user's location changes, keeping the location dot in the center.

because the center is recorded:

> mapCenterLat: the latitude of the center of the current map view
> mapCenterLng: the longitude of the center of the current map view

Is it really necessary to record these attributes?

scottsfarley93:
@pathmapper While I think these properties would be helpful in improving relevance scores for some searches, I understand the concern.  I removed the the `mapCenterLat` and `mapCenterLng` properties from the event.

andrewharvey:
Worth noting if `trackProximity` is enabled it will still leak the mapCenter via proximity. In my opinion I think it's fine, and would be valuable to have that information to help improve the geocoder.

Issue #188, opened 2019-02-26
+1MB of JS - the unpleasant effect of anonymous event tracking


jimutt:
Hi!
I recently discovered that the 3.1.2 release (https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v3.1.2/mapbox-gl-geocoder.min.js) has a size of over 1MB. Compared to the modestly sized 3.1.1 release at 36k: https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-geocoder/v3.1.1/mapbox-gl-geocoder.min.js

I believe it maybe was an unfortunate mistake to include dependencies (Without any investigation I guess the `request` and `crypto` packages are the main cause) inducing this huge increase in size.

andrewharvey:
wow good catch @jimutt.

It does indeed appear to be due to dependencies introduced in #183, @scottsfarley93 hope you're able to take a look.

scottsfarley93:
Interesting. I can take a look at this later this week.

scottsfarley93:
After removing dependencies, the js bundle is now 75.148kb. It is available in v3.1.6.

PR #189, opened 2019-02-27
Remove dependencies for events logging

scottsfarley93:
In response to #188, this PR investigates the increase in bundle size introduced in #183 and attempts to reduce the increase by changing the dependencies used in logging interaction events.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var https = require('https');
-var crypto = require('crypto');
+var shajs = require('sha.js')
 
 /**
  * Construct a new mapbox event client to send interaction events to the mapbox event service
@@ -182,9 +182,7 @@ MapboxEventManager.prototype = {
    * @private
    */
   generateSessionID: function () {
-    var sha = crypto.createHash('sha256');
-    sha.update(Math.random().toString());
-    return sha.digest('hex')
+    return new shajs.sha256().update(Math.random().toString()).digest('hex')
   },
 
   /**
--- a/lib/events.js
+++ b/lib/events.js
@@ -215,7 +215,7 @@ MapboxEventManager.prototype = {
    * @private
    */
   shouldEnableLogging: function (options) {
-    if (!this.origin.includes('api.mapbox.com')) return false;
+    if (this.origin.indexOf('api.mapbox.com')  == -1) return false;
     // hard to make sense of events when a local instance is suplementing results from origin
     if (options.localGeocoder) return false;
     // hard to make sense of events when a custom filter is in use
--- a/lib/events.js
+++ b/lib/events.js
@@ -1,8 +1,5 @@
 'use strict';
-
-var https = require('https');
 var shajs = require('sha.js')
-
 /**
  * Construct a new mapbox event client to send interaction events to the mapbox event service
  * @param {Object} options options with which to create the service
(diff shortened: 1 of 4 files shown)


andrewharvey:
For comparison

```
v3.1.1: 37K
v3.1.2: 1.1M
this PR:131K
```

Not sure what the split of https vs sha.js is but for reference it looks like mapbox-sdk-js and mapbox-gl-js both use XMLHttpRequest without any overhead.

https://github.com/mapbox/mapbox-sdk-js/blob/master/lib/browser/browser-layer.js
https://github.com/mapbox/mapbox-gl-js/blob/master/src/util/ajax.js 

Wondering if it's worth doing the same here.

scottsfarley93:
Using XMLHttpRequest, the min.js file size is now 75132 bytes.

lib/events.js
```diff
@@ -93,121 +91,131 @@ MapboxEventManager.prototype = {
...
-     * @private
-     */
-  shouldEnableLogging: function(options){
-    if (!this.origin.includes('api.mapbox.com')) return false;
+   * Check whether events should be logged
+   * Clients using a localGeocoder or an origin other than mapbox should not have events logged
+   * @private
+   */
+  shouldEnableLogging: function (options) {
+    if (this.origin.indexOf('api.mapbox.com')  == -1) return false;
```
andrewharvey: Would `===` bet safer? It's what's used at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf#Checking_occurrences

andrewharvey:
@scottsfarley93 Thanks for fixing these things up.

I wanted to make some comments about the project/development in general. It's not for me to dictate how things are done, but I've been finding it hard to keep tabs on development and maintenance of this project. I'd prefer if we could:

1.Try to keep each PR contained to one issue, it's easier to review and keep track of where changes are being made.

2. Follow (or update) the deploy instructions https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CONTRIBUTING.md#deploying and have a commit for each release (instead of bundling releases as part of other code changes). At the moment it's hard to see where releases are happening because they are mixed in with other commits.

3. Make a PR for each release, this gives an opportunity for a review of the release before it's pushed out, to hopefully catch any potential issues beforehand.

4. If significant changes and new features are being released, use a minor version bump instead of a patch release, following semver (for example, in my opinion the events logging was too large a change for a patch release).

scottsfarley93:
@andrewharvey thanks for the review and the contribution feedback. Quick question there -- I'm not sure I follow the order of operations on the [deployment instructions](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CONTRIBUTING.md#deploying). Specifically, when does the merge to master come in?

Based on your suggestion, I think the following order makes the most sense, but wanted to double check:
1. Make feature changes/bug fixes in PR
2. Merge PR to master
3. Update changelog and version in new PR
4. Merge changelog/version PR into master
5. Tag release and npm publish

Let me know what you're thinking there. I can update the docs accordingly once we've nailed down the steps.

andrewharvey:
@scottsfarley93 That's right, I find that approach, forcing releases to have their own commits rather than doing a release together with other commits, a lot clearer and better practice and what has historically been happening. #198 clarifies that.

PR #191, opened 2019-02-27
Change dead url

n15e:
The url provided for the options.types description is wrong. Replacing with a working link.


andrewharvey:
hey @n15e thanks for the PR. The API.md is auto generated from the source, could you update it in the source https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L33 then run `npm run docs` and commit that?

PR #193, opened 2019-02-28
 Change dead url

n15e:
The url provided for the options.types description is wrong. Replacing with a working link. Resubmitting PR as mentioned on #191 . Not entirely sure how to do this without cancelling the older PR.


tristen:
Thanks!

Issue #195, opened 2019-03-13
Improve default options set to improve search relevance


scottsfarley93:
Looking at the options that are set on the geocoding call (https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L58), there are several defaults that could be configured to provide a better out-of-the box search experience. 

Specifically, I think good improvements could be found with:
- setting `trackProximity` to default true -- searches with a user-relevant proximity are likely to be more relevant
- setting a default language based on the user's browser settings

I suspect this may change user's expectations and/or be a breaking change -- so we could target this for the next major release.

andrewharvey:
I think these are positive changes.

> I suspect this may change user's expectations and/or be a breaking change -- so we could target this for the next major release.

I agree.

katydecorah:
👍 I also think that these are positive changes

andrewharvey:
Done in https://github.com/mapbox/mapbox-gl-geocoder/tree/version4 by #200 #202, slated for v4 release.

Issue #150, opened 2018-04-13
default placeholder text should be localized if language option provided


vfoss:
The placeholder text in the search bar for the geocoder remains "Search" (in English), even when the language of the map is changed, as in 
`let geocoder = new MapboxGeocoder({accessToken: mapboxgl.accessToken, language: 'es' });`.  I cannot figure out a clean way to change it.

andrewharvey:
@vfoss You can set it with the [`placeholder` option](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md#mapboxgeocoder).

But let's leave this ticked open to make the default placeholder set by the first language if provided.

andrewharvey:
Done in https://github.com/mapbox/mapbox-gl-geocoder/tree/version4 by #201, slated for v4 release.

PR #200, opened 2019-03-13
Obtain language from browser by default

scottsfarley93:
As per #195, this PR checks the browser's `language` tag and uses this by default when making searches if another language is not already set in the constructor options.

This also unblocks #150, which we can handle next.

\cc @katydecorah @kbauhaus @yuletide

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,7 @@ var exceptions = require('./exceptions');
 var MapboxClient = require('@mapbox/mapbox-sdk');
 var mbxGeocoder = require('@mapbox/mapbox-sdk/services/geocoding');
 var MapboxEventManager = require('./events');
+// var subtag = require('subtag');
 var geocoderService;
 
 /**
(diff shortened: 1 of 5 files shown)


lib/index.js
```diff
@@ -388,6 +391,17 @@ MapboxGeocoder.prototype = {
     return this.options.proximity;
   },
 
+  /**
+   * Get the language to use in UI elements and when making search requests
+   * 
+   * Look first at the explicitly set options otherwise use the browser's language settings
+   */
```
andrewharvey: For completeness given this is a public method, I think we should add JSDoc for `@returns`.

lib/index.js
```diff
@@ -388,6 +391,19 @@ MapboxGeocoder.prototype = {
...
+  /**
+   * Get the language to use in UI elements and when making search requests
+   * 
+   * Look first at the explicitly set options otherwise use the browser's language settings
+   * 
+   * @returns {String} The language used by the geocoder
+   */
+  getLanguage: function(){
+    var browserLocale = navigator.language || navigator.userLanguage || navigator.browserLanguage;
+    console.log(browserLocale);
```
katydecorah: Remove this console.log()?

scottsfarley93: 👍 removed.

PR #198, opened 2019-03-13
Documentation suggestions for the deploy, PR process

katydecorah:
I'd like to propose some documentation updates to the deploy process. The goal is to help make it clear for contributors what to include in their PR and how to deploy. In this PR, I propose:

* Step-by-step deployment process to help standardize the release process.
* Encourage consistent updating to changelog with un-released changes under a `master` heading.
* Add pull request template with small checklist to help remind contributor of tasks (provide details, write tests, update changelog).

@andrewharvey I'd love to hear your feedback on these proposed updates 🙏

PR #201, opened 2019-03-13
Localize placeholder based on first language

scottsfarley93:
Implements the change suggested in #150 by localizing the placeholder value based on the provided language option set in the constructor.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -8,6 +8,8 @@ var exceptions = require('./exceptions');
 var MapboxClient = require('@mapbox/mapbox-sdk');
 var mbxGeocoder = require('@mapbox/mapbox-sdk/services/geocoding');
 var MapboxEventManager = require('./events');
+var localization = require('./localization');
+var subtag = require('subtag');
 var geocoderService;
 
 /**
@@ -56,7 +58,6 @@ function MapboxGeocoder(options) {
 
 MapboxGeocoder.prototype = {
   options: {
-    placeholder: 'Search',
     zoom: 16,
     flyTo: true,
     trackProximity: false,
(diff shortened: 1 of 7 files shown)


lib/index.js
```diff
@@ -388,6 +389,24 @@ MapboxGeocoder.prototype = {
...
   },
 
+  /**
+   * Get the text to use as the search bar placeholder
+   * 
+   * If placeholder is provided in options, then use options.placeholder
+   * Otherwise, if language is provided in options, then use the localized string of the first language if available
+   * Otherwise use the default
+   */
+  getPlaceholderText: function(){
```
andrewharvey: should this method be private?

lib/index.js
```diff
@@ -388,6 +389,24 @@ MapboxGeocoder.prototype = {
...
+    if (this.options.placeholder) return this.options.placeholder;
+    if (this.options.language){
+      var firstLanguage = this.options.language.split(",")[0];
+      var language = subtag.language(firstLanguage);
+      var localizedValue = localization.placeholder[language];
+      if (localizedValue)  return localizedValue;
+    }
+    return 'Search';
```
andrewharvey: you could also lookup 'en' in localization.placeholder, but I don't think it really matters.

lib/index.js
```diff
@@ -388,6 +389,24 @@ MapboxGeocoder.prototype = {
     return this.options.proximity;
   },
 
+  /**
+   * Get the text to use as the search bar placeholder
+   * 
+   * If placeholder is provided in options, then use options.placeholder
+   * Otherwise, if language is provided in options, then use the localized string of the first language if available
+   * Otherwise use the default
```
andrewharvey: I think we should add a JSDoc `@returns` here for completeness.

lib/index.js
andrewharvey: related to https://github.com/mapbox/mapbox-gl-geocoder/pull/198/files#r265396617 but I think it's better to re-gen the docs with each commit and include them.

scottsfarley93: Good catch. Moved this to be a private method.

Issue #153, opened 2018-05-29
include matching text in suggestions


andrewharvey:
Currently results will be displayed in whichever language was set in the options.

`{ language: 'en', country: 'au' }`

![en_au](https://user-images.githubusercontent.com/117278/40681595-eb112c12-63cc-11e8-8c5f-b23873bf905b.png)

`{ language: 'zh', country: 'au' }`

![zh_au](https://user-images.githubusercontent.com/117278/40681657-13fca322-63cd-11e8-9435-9fe656b176cf.png)

To get results displayed in one language, but still display the matching text if the query was in another language you need to do something like:

```
geocoder._typeahead.getItemValue = function (item) {
    if (item.matching_text) { // matching text is the text the query matched, so will be in the same language that the user queried in
        return item.matching_text + ' (' + item.place_name + ')';
    } else {
        return item.place_name;
    }
};
```

![zh_au_matching_text](https://user-images.githubusercontent.com/117278/40681878-aabb2252-63cd-11e8-813e-0ea2d32f1c02.png)

That feels a bit hacky though, I'm thinking if doing this be better supported without overwriting the internal `_typeahead.getItemValue` method, or even if this should become the default?

scottsfarley93:
What about adding a constructor option to provide a function that allows implementation-specific customization over how the results are displayed? 

Something like:

```
var geocoder = new MapboxGeocoder({ 
    accessToken: mapboxgl.accessToken,
    render: function(result){
        return item.matching_text + ' (' + item.place_name + ')'
   }
})
```

We could provide a sensible default (either keep it the same as it is now (use `place_name`)  or use a function that renders the matching text in the dropdown), but provide the ability to override with something different if desired.

andrewharvey:
:+1: I think that's the way to go.

Rethinking about this, the current defaults are probably good as they are.

scottsfarley93:
The ability to do this was introduced in #208. While the defaults have not changed, once version 4 is released, additional fields can be accessed via the `render` function.

PR #208, opened 2019-03-15
Custom html render

scottsfarley93:
This PR approaches #153 by exposing two new constructor options `render` and `getItemValue`. 

-  `render` is a function that accepts a carmen geojson function and returns an html string to display in the dropdown menu. This function can be customized by setting `options.render` so that any html can be rendered in the dropdown. This improves the customizability of the style/display of the results. 
-  `getItemValue` is a function that accepts a carmen geojson feature and returns a string to use in the search bar after a selection has been made. This method currently retains the old name, but perhaps we should name it something more intuitive in this context. 

Note: This requires the changes in https://github.com/tristen/suggestions/pull/25 -- so we'll wait here until this upstream PR lands.

\cc @katydecorah @yuletide @kbauhaus


lib/index.js
```diff
@@ -63,7 +63,16 @@ MapboxGeocoder.prototype = {
...
+    origin: 'https://api.mapbox.com',
+    render: function(feature){
+      var maki;
+      if (feature.properties.maki) maki = feature.properties.maki;
+      else maki = 'marker'
+      return `<img src="https://unpkg.com/@mapbox/maki@6.1.0/icons/${maki}-11.svg"/> ${feature.place_name}`;
+    },
+    getItemValue: function(item) { 
+      return item.place_name
+    }
```
andrewharvey: Is this just for your testing, or are you intending to default to showing an icon?

scottsfarley93: Still just a WIP -- I'm planning on changing this back to match the existing functionality before merge. 

It might be helpful to use this as an example [here](https://docs.mapbox.com/mapbox-gl-js/examples/#geocoder) since it would demonstrate the custom render function and be a use case I suspect people are pretty familiar with.

Issue #167, opened 2018-09-27
Minor styling suggestion when dropdown is open


jdaudier:
When the dropdown is open for the geocoder, the `mapboxgl-ctrl-geocoder` div should have `0` bottom border radius. Minor styling suggestion.

## BEFORE
![](https://zappy.zapier.com/35C2FDB7-1FE1-4C0A-90BB-1529288D17DF.png)

## AFTER
![](https://zappy.zapier.com/FC68C066-0BF5-4104-81AD-5283DDB534E5.png)

katydecorah:
Thank you for reporting! This fix was completed in #211 and will be live in version 4 #197

PR #211, opened 2019-03-15
[v4] style updates

### LINKED ISSUE #209: [v4] update input and results styles
Tracking the design updates we’ll be making in v4 #197  (This ticket is not tracking the mobile styles #89 , we’ll accomplish after we come to a consensus on the design updates made here.)

The goal is to bring the UI up-to-date with Assembly, while maintaining minimal styling so developers can readily theme the controls.

Current | Proposed
--------|-----------
image |  image

### Proposed updates to: Input

* [ ] Use Assembly icons for search 🔍, close ❌, and loader ♻️
    * [ ] run icons through svgo to squeeze out excess characters (possible this was already done in Assembly)
    * [ ] 🔍 make the same color as the input font color
    * [ ] ❌ add a hover state
    * [ ] ❌ consider making it display only when the user hovers over the input box, like how the native search input works https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
    * [ ] ♻️ add the same animation (spin) as Assembly
* [ ] Add “Open Sans” to font stack
* [ ] Increase input font-size to at least 15px to improve legibility
* [ ] Adjust input padding to match Assembly `6px 12px`
* [ ] Use border-radius `4px` to match Assembly, this is also the same border-radius used by GL buttons

### Proposed updates to: Results

* [ ] Increase font-size to at least 13px to improve legibility
* [ ] Adjust padding to match Assembly `6px 12px` **or** add more y padding and remove border-bottom.
* [ ] Add `3px` or `6px` between the bottom of input and top of search results **or** turn off border-bottom on input when displaying results to join the input and the results seamlessly (ref #167).

These changes will be light, but barring other suggestions, I think that's ok. I want to be mindful that the results don't take up too much more space than they currently do and are still theme-able.

### Styles that bump against other GL controls

* [ ] Relax box-shadow so it provides depth without giving the appearance of a transparent border. Other GL controls have the same box-shadow. Should we consider updating their box-shadow to match? I updated the box-shadow on the GL controls in the Propose screenshot.

### Questions

* Am I missing any other important GL controls that we should compare against? 
* Should we provide a dark theme?

Ready for your thoughts and happy to take suggestions at any questions above.

cc @samanpwbb @kbauhaus @scottsfarley93 @yuletide

katydecorah:

This PR brings the UI up-to-date with Assembly, while maintaining minimal styling so developers can readily theme the controls fixes #209 fixes https://github.com/mapbox/mapbox-gl-geocoder/issues/167

### To do
Input:
* [x] Use Assembly icons for search 🔍, close ❌, and loader ♻️
    * [x] run icons through svgo to squeeze out excess characters (possible this was already done in Assembly)
    * [x] 🔍 make the same color as the input font color
    * [x] ❌ add a hover state
    * [x] ❌ consider making it display only when the user hovers over the input box, like how the native search input works https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
    * [x] ♻️ add the same animation (spin) as Assembly
* [x] Add “Open Sans” to font stack
* [x] Increase input font-size to at least 15px to improve legibility
* [x] Adjust input padding to match Assembly `6px 12px`
* [x] Use border-radius `4px` to match Assembly, this is also the same border-radius used by GL buttons

Results:

* [x] Increase font-size to at least 13px to improve legibility
* [x] Adjust padding to match Assembly `6px 12px` **or** add more y padding and remove border-bottom.
* [x] Add `3px` or `6px` between the bottom of input and top of search results **or** turn off border-bottom on input when displaying results to join the input and the results seamlessly (ref #167).
* [x] Should we present result in more than one line **or** continue with single line?

Clean up:
* [x] Double check CSS to make sure every line is in use

@kbauhaus I was thinking we could work on styling the results during our pairing session today


lib/mapbox-gl-geocoder.css
```diff
@@ -3,137 +3,174 @@
...
+}
+
+.geocoder-icon-search {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='18' height='18'%3E%3Cpath d='M7.4 2.5c-2.7 0-4.9 2.2-4.9 4.9s2.2 4.9 4.9 4.9c1 0 1.8-.2 2.5-.8l3.7 3.7c.2.2.4.3.8.3.7 0 1.1-.4 1.1-1.1 0-.3-.1-.5-.3-.8L11.4 10c.4-.8.8-1.6.8-2.5.1-2.8-2.1-5-4.8-5zm0 1.6c1.8 0 3.2 1.4 3.2 3.2s-1.4 3.2-3.2 3.2-3.3-1.3-3.3-3.1 1.4-3.3 3.3-3.3z' fill='%23757575'/%3E%3C/svg%3E");
+}
+
+.geocoder-icon-close {
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='8' height='8'%3E%3Cpath d='M.8 0C.4 0 0 .4 0 .8c0 .2.1.4.3.5L2.9 4 .3 6.6c-.2.2-.3.4-.3.6 0 .4.4.8.8.8.2 0 .4-.1.6-.3L4 5.1l2.6 2.6c.1.2.4.2.6.2.4 0 .8-.4.8-.8 0-.2-.1-.4-.2-.6L5.1 4l2.6-2.7c.2-.1.3-.3.3-.5 0-.4-.4-.8-.8-.8-.2 0-.4.1-.6.2L4 2.8 1.4.2C1.2.1 1 0 .8 0z' fill='%23fff'/%3E%3C/svg%3E");
+  display: none;
+}
```
samanpwbb: Could we implement the svg icons in such a way that their properties are style-able with CSS (for example, by inlining the CSS)?

lib/mapbox-gl-geocoder.css
```diff
@@ -3,137 +3,174 @@
...
+  right: 10px;
+  top: 10px;
+  display: none;
+}
 
 .mapboxgl-ctrl-geocoder,
 .mapboxgl-ctrl-geocoder ul {
-  box-shadow: 0 0 0 2px rgba(0,0,0,0.1);
-  }
+  box-shadow: 0 3px 5px rgba(0, 0, 0, 0.15), 0 -1px 0 rgba(0, 0, 0, 0.02);
```
samanpwbb: The shadow looks nice, but all assembly's shadows are perfectly centered and not vertically offset. I get that offset shadows look more natural, but, for consistency, we'd probably override this run in studio. I wonder if it'd be worth filing an issue in Assembly about offsetting all shadows by default.

lib/index.js
```diff
@@ -166,10 +170,18 @@ MapboxGeocoder.prototype = {
     }
   }, 200),
 
+  _showButton: function() {
+    if (this._typeahead.selected) this._clearEl.style.display = 'block';
+  },
```
samanpwbb: When close button gets added, we should add some extra right padding to the input to shift where truncation begins:
Screen Shot 2019-03-19 at 11 34 44 AM

lib/mapbox-gl-geocoder.css
```diff
@@ -3,137 +3,174 @@
...
+.mapboxgl-ctrl-geocoder .geocoder-icon-search {
+  position: absolute;
+  top: 10px;
+  left: 9px;
+}
+
+.mapboxgl-ctrl-geocoder .geocoder-icon-close {
+  padding: 0;
+  margin: 0;
+  background-color: #757575;
```
samanpwbb: The style of this button doesn't conform to existing UI patterns in Studio. I understand the desire to differentiate the close action from the magnifying glass, but I think we'd want to drop the outline and change the fill color of the X when adding the geocoder to studio or consistency.

lib/mapbox-gl-geocoder.css
```diff
@@ -3,137 +3,174 @@
...
-  font:15px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
-  position:relative;
-  background-color:white;
-  width:33.3333%;
-  min-width:240px;
-  max-width:360px;
-  z-index:1;
-  border-radius:3px;
-  }
+  font: 15px/20px "Open Sans", "Helvetica Neue", Arial, Helvetica, sans-serif;
```
samanpwbb: I'm seeing system font in the input box and open sans in the results list:
Screen Shot 2019-03-19 at 11 39 12 AM

katydecorah:
@samanpwbb thanks for the review! I believe I've resolved all your suggestions 🙇‍♀️ 

> While running this PR I noticed that I can no longer tab or arrow through results. Is this a known issue

Yes! It will be fixed in #220

scottsfarley93:
Changes look good to me! Overall I really like new look/feel. 

Not sure if it's in scope for this PR, but I do find it's quite hard to read long-named place names. I see `Should we present result in more than one line or continue with single line?` on the outline of changes above. If we don't make a change in this PR to improve legibility of many-character names, let's open a ticket and tackle it later. 

Screen Shot 2019-03-20 at 1 28 42 PM

samanpwbb:
> If we don't make a change in this PR to improve legibility of many-character names, let's open a ticket and tackle it later.

Studio uses the primary name on one line and then full address on the preceding line. From a legibility POV I do this this would be best:

Screen Shot 2019-03-20 at 2 04 32 PM

katydecorah:
Multiline is the way to go 👍 I think it makes sense to use the custom work in #208 to set a default `render` as multiline text. I'll finish up that change in this PR.

katydecorah:
Latest commit adds a default render function to create multiline results:

image

andrewharvey:
looking very nice @katydecorah .

How about setting the css to end overflowing text with ..., it's visually a bit nicer and tells the user some text is hidden.

katydecorah:
image

> setting the css to end overflowing text with ..

@andrewharvey good catch! Added ✅

andrewharvey:
@katydecorah Just checking, is there meant to be this grey box on the clear button?

Screenshot from 2019-03-22 12-11-52

I also noticed that the loading circle will overlap the clear x, but it's not a big deal.

katydecorah:
@andrewharvey oh shoot, thanks for flagging. I'll get that fixed in another PR 🙇‍♀️

PR #212, opened 2019-03-15
Event logging update

scottsfarley93:
This PR updates the interaction logging first introduced in #183 in the following ways:
- `place_name` and `id` are included with selection events so that these events can be more easily consumed by search engine updates
- a new `keyevent` event is introduced. This event will be sent on each keydown event -- these anonymous events will be used to further improve relevance computation in the core search engine.
 - events are batched together to minimize network calls


lib/events.js
```diff
@@ -40,31 +46,49 @@ MapboxEventManager.prototype = {
...
+   * @private
+   * @param {Object} keyEvent the keydown event to log
+   * @param {Obeject} geocoder a mapbox-gl-geocoder instance
+   * 
+   */
+  keyevent: function(keyEvent, geocoder){
+    //pass invalid event
+    if (!keyEvent.key) return;
+    // don't send events for keys that don't change the input
+    if (keyEvent.metaKey || [9, 27, 37, 39, 13, 38, 40].indexOf(keyEvent.keyCode) !== -1) return;
```
ingalls: Can you add a comment with the actual keys that these codes correspond to

lib/events.js
```diff
@@ -82,9 +106,10 @@ MapboxEventManager.prototype = {
       return callback();
     }
     var options = this.getRequestOptions(payload);
+    var _this = this;
     this.request(options, function(err,res){
-      if (err) return this.handleError(err, callback);
-      if (res.statusCode != 204) return this.handleError(res, callback);
+      if (err) return _this.handleError(err, callback);
+      if (res.statusCode != 204) return _this.handleError(res, callback);
       if (callback) return callback();
```
ingalls: This is duplicative of line 104 in that `callback` is populated if it is falsy. Line 104 should be removed.

lib/events.js
```diff
@@ -82,9 +106,10 @@ MapboxEventManager.prototype = {
       return callback();
     }
     var options = this.getRequestOptions(payload);
+    var _this = this;
     this.request(options, function(err,res){
-      if (err) return this.handleError(err, callback);
-      if (res.statusCode != 204) return this.handleError(res, callback);
+      if (err) return _this.handleError(err, callback);
+      if (res.statusCode != 204) return _this.handleError(res, callback);
```
ingalls: Are we able to use arrow functions? If possible let's switch this to an arrow function to avoid the rescope of `this` and avoid having to use `_this`

lib/events.js
```diff
@@ -207,12 +234,49 @@ MapboxEventManager.prototype = {
...
     if (options.filter) return false;
     return true;
+  },
+
+  /**
+   * Flush out the event queue by sending events to the events service
+   * @param {Function} callback the function to execute after the events have been sent successfully
+   * @private
+   */
+  flush: function(){
```
ingalls: The JS Doc says this function should have a `callback` param

lib/events.js
scottsfarley93: No arrow functions here just now -- perhaps in a future sprint we can bring support for arrow functions. 

In lieu of arrow functions, however, I added a `bind(this)` so we don't need to create a temporary `_this` variable.

ingalls:
:taco:

PR #217, opened 2019-03-18
No hardcoded ids

scottsfarley93:
The mapbox geocoder does not guarantee stability of the `id` property on returned feature. In order to ensure that the exceptions are applied even if the underlying IDs change, this PR changes the bounding box exceptions checking to use the `properties.short_code` on returned features, which specifies a ISO 3166-1 country code which provides a more stable identifier to check against. 

\cc @katydecorah @yuletide


lib/index.js
```diff
@@ -172,17 +172,17 @@ MapboxGeocoder.prototype = {
     var selected = this._typeahead.selected;
     if (selected) {
       if (this.options.flyTo) {
-        if (!exceptions[selected.id] && selected.bbox) {
+        if (!exceptions[selected.properties.short_code] && selected.bbox) {
```
katydecorah: @scottsfarley93 is it possible for a result to not have `properties` ? I want to check because if a result didn't have `properties` then this would throw an error. If there is a doubt, we could update this (and similarly on line 178) to:

```suggestion
        if (selected.properties && !exceptions[selected.properties.short_code] && selected.bbox) {
```

PR #218, opened 2019-03-18
Deduplicate result event emissions (#99)

scottsfarley93:
This PR addresses the bug reported in #99 by keeping a state variable indicating the last selected feature. 

\cc @katydecorah @yuletide

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -54,6 +54,7 @@ function MapboxGeocoder(options) {
   this.options = extend({}, this.options, options);
   this.inputString = '';
   this.fresh = true;
+  this.lastSelected = null;
 }
 
 MapboxGeocoder.prototype = {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -190,8 +191,12 @@ MapboxGeocoder.prototype = {
           });
         }
       }
-      this._eventEmitter.emit('result', { result: selected });
-      this.eventManager.select(selected, this);
+      if (selected.id !== this.lastSelected){
+        console.log("gl:result")
```
katydecorah: Remove this console.log()?

scottsfarley93: Good catch!

andrewharvey:
I must admit, this is an ingenious workaround for the underlying bug in suggestions. :wink: Certainly an improvement over the status quo.

There is a corner case where you re-select the same result and the result event is skipped, but I think that's too minor and arguably shouldn't re-fire anyway, that it's okay.

Issue #213, opened 2019-03-15
Add marker to map showing result


scottsfarley93:
It's a little jarring right now to select a search result, have the map fly to those coordinates, and not see anything on the map. It could be a good idea to add a marker to the map indicating the location of the feature that has been returned. We could enable this as an optional feature so those users who manage a custom display of the result could continue to do so. 

\cc @katydecorah

andrewharvey:
I agree with making this optional, since some developers would be implementing this already and may like to continue to do so. However I see this as providing better defaults out of the box, so :+1:

andrewharvey:
We'll need to decide what to do with https://docs.mapbox.com/mapbox-gl-js/example/point-from-geocoder-result/ in light of this PR.

scottsfarley93:
This was implemented in #219 and will the be default behavior once version 4 is released.

A key note about this change is that [`mapboxgl`](https://github.com/mapbox/mapbox-gl-js) is now a peer dependency of this package. New mapbox-gl-geocoder instances should set `mapboxgl` in the constructor options.

If `mapboxgl` is not found in the constructor options, markers will not be added to the map and an error will be generated in the console, but the plugin will otherwise function as usual.

PR #219, opened 2019-03-19
Map marker on selected result

scottsfarley93:
This PR handles #213 by adding a mapboxgl#Marker to the map after the user has selected a result. it exposes two additional constructor options:

- `addToMap`: a boolean field indicating whether the marker should be added. Optional, default `true`. 
- `markerOptions`: an optional object specify the [marker options](https://docs.mapbox.com/mapbox-gl-js/api/#marker) to be passed to the marker constructor. 

**Note**: this requires a more recent version of mapbox-gl-js (v0.53.0) so there may be compatibility issues. Also, the existing tests need some massaging to get to work with the newer version of mapbox-gl. 

- [x] Make sure tests work with more recent mapbox-gl-js version

\cc @katydecorah @yuletide


scottsfarley93:
This bumps the bundle size up a nontrivial amount:
```
5194 mapbox-gl-geocoder.css
711473 mapbox-gl-geocoder.min.js
```

We should make sure we're comfortable with such an increase before merging. 

\cc @andrewharvey -- what are your thoughts on about the tradeoff here?

andrewharvey:
> This bumps the bundle size up a nontrivial amount:

Is that because it's now including the whole mapbox-gl in the geocoder bundle? If that's the case, I think we should try to find a way avoid that.

samanpwbb:
To get around an explicit mapboxgl dependency, you could set mapbox gl as a peer dependency in package.json, and make the geocoder take mapboxgl as an argument on initialization.

API.md
```diff
@@ -44,6 +44,8 @@ A geocoder component using Mapbox Geocoding API
     -   `options.localGeocoder` **[Function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/function)?** A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
     -   `options.reverseMode` **(`"distance"` \| `"score"`)** Set the factors that are used to sort nearby results. (optional, default `'distance'`)
     -   `options.reverseGeocode` **[boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)?** Enable reverse geocoding. Defaults to false. Expects coordinates to be lat, lon.
+    -   `options.addToMap` **[Boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** Indicates that the location of the user-selected result should be added to the map as a [mapbox-gl#Marker](https://docs.mapbox.com/mapbox-gl-js/api/#marker) (optional, default `true`)
```
andrewharvey: Do you think naming this something like `showMarker` would be clearer by indicating this specifically relates to a marker? `addToMap` at first glance to me means add the geocolate control to the map, which is done automatically already.

Do you think it's clearer like this? That's the general style used commonly in GL JS docs, for example https://docs.mapbox.com/mapbox-gl-js/api/#popup

    "If `true`, a [Marker](...) will be added to the map at the location of the user-selected result."

API.md
```diff
@@ -44,6 +44,8 @@ A geocoder component using Mapbox Geocoding API
...
+    -   `options.markerOptions` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)?** An object specifying the properties to use when constructing the marker, if `addToMap` is true
```
andrewharvey: We should link to https://docs.mapbox.com/mapbox-gl-js/api/#marker for easy access to available options.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Localize placeholder based on language set in constructor options [#150](https://github.com/mapbox/mapbox-gl-geocoder/issues/150)
 - `trackProximity` turned on by default [#195](https://github.com/mapbox/mapbox-gl-geocoder/issues/195)
 - Bump suggestions to v1.3.4
+- Add constructor properties that allow adding the selected result to the map as a mapbox-gl#marker. Adding the result to the map is now the default behavior.
```
andrewharvey: Something I've personally found helpful in CHANGELOG's is separating out breaking changes, features, and bug fixes, eg https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#0490. This makes it easier to see the most important changes like breaking changes.

andrewharvey:
I haven't tested, so not what the current behaviour is, but if someone clears the geocoder, should the marker be removed? I'm think, yes it should be removed, but what do you think?

andrewharvey:
I wonder should this work like the new flyTo option, where you can pass true, false, or an options object? Simplifying the two options into a single `marker` option which can be true, false or an options object?

scottsfarley93:
Thanks for the feedback @andrewharvey!

> if someone clears the geocoder, should the marker be removed?
Good catch.  I agree, and added this behavior in https://github.com/mapbox/mapbox-gl-geocoder/pull/219/commits/f454c4d1bab66b0f816493b192239f324d7e1355.

> Simplifying the two options into a single marker option which can be true, false or an options object
I think this simplifies the API and, as you point out, makes the behavior more like the existing options.

CHANGELOG.md
scottsfarley93: This is a great example and one we can emulate going forward.

package.json
```diff
@@ -47,17 +47,19 @@
     "insert-css": "2.0.0",
     "lint-staged": "^8.1.5",
     "lodash.once": "^4.0.0",
-    "mapbox-gl": "^0.27.0",
+    "mapbox-gl": "^0.53.0",
```
katydecorah: To confirm, do we want mapbox-gl as a devDep and peerDep? I'm not well versed on how peerDep work so just want to gut check in case this was in error.

lib/index.js
```diff
@@ -42,6 +42,7 @@ var geocoderService;
  * @param {Function} [options.localGeocoder] A function accepting the query string which performs local geocoding to supplement results from the Mapbox Geocoding API. Expected to return an Array of GeoJSON Features in the [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) format.
  * @param {'distance'|'score'} [options.reverseMode='distance'] - Set the factors that are used to sort nearby results.
  * @param {boolean} [options.reverseGeocode] Enable reverse geocoding. Defaults to false. Expects coordinates to be lat, lon.
+ * @param {Boolean|Object} [options.marker=true]  If `true`, a [Marker](https://docs.mapbox.com/mapbox-gl-js/api/#marker) will be added to the map at the location of the user-selected result using a default set of Marker properties.  If the value is an object, the marker will be constructed using these properties. If `false`, no marker will be added to the map.
```
andrewharvey: sorry to nit pick on you but https://docs.mapbox.com/mapbox-gl-js/api/#marker uses marker options not marker properties. Being consistent with terminology makes it easier for newcomers. So I prefer we use "options" rather than "properties" here.

lib/index.js
```diff
@@ -65,7 +66,9 @@ MapboxGeocoder.prototype = {
     minLength: 2,
     reverseGeocode: false,
     limit: 5,
-    origin: 'https://api.mapbox.com'
+    origin: 'https://api.mapbox.com',
+    marker: true,
+    mapboxgl: null
```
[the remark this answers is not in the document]
andrewharvey: I agree, we should document this option.

lib/index.js
```diff
@@ -133,6 +136,15 @@ MapboxGeocoder.prototype = {
       this._map.on('moveend', this._updateProximity);
     }
 
+    this.mapMarker = null;
+    this._handleMarker = this._handleMarker.bind(this);
+
+    this._mapboxgl = this.options.mapboxgl;
+    if (!this._mapboxgl) {
+      console.error("No mapboxgl detected. Map markers are disabled. Please set options.mapboxgl.");
```
andrewharvey: Should we only log a console error if this.marker is not null? After all, if we aren't setting markers, then we don't need the options.mapboxgl so it's not actually an error.

package.json
scottsfarley93: I believe we want it to be both. During regular use, we want it as a peer dependency because we expect the plugin to be used in conjunction with a mapbox-gl map. However, during development we want it as a dev dependency so we can use it in tests.

lib/index.js
scottsfarley93: 👍 Thanks for the thoughtful & thorough feedback! Consistency is good. Changing to `options`.

lib/index.js
```diff
@@ -503,6 +522,37 @@ MapboxGeocoder.prototype = {
...
+   * Handle the placement of a result marking the selected result
+   * @private
+   * @param {Object} selected the selected geojson feature
+   * @returns {MapboxGeocoder} this
+   */
+  _handleMarker: function(selected){
+    // clean up any old marker that might be present
+    this._removeMarker();
+    var defaultMarkerOptions = {
+      color: '#4668F2'
```
andrewharvey: Since this feature was only added to mapbox-gl-js in v0.45.0, I think we should add a note to the changelog about the new minimum gl-js version requirement for mapbox-gl-geocoder v4, like was done for v2 https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CHANGELOG.md#v200

package.json
andrewharvey: Is v0.53.0 the minimum requirement? I think if possible we should try to set this to the minimum version required for everything to work in gl-geocoder.

Not essential, just a nice to have.

jayenashar:
this is working pretty well for me, but how do i get it to move an existing marker, that i'm using as a default before the user enters a location in the geocoder?  can i pass the `marker` option an existing marker?

Issue #162, opened 2018-08-14
Trapped focus causes accessibility issues


3zzy:
Looks like this line:

```
// TAB, ESC, LEFT, RIGHT, ENTER, UP, DOWN
if (e.metaKey || [9, 27, 37, 39, 13, 38, 40].indexOf(e.keyCode) !== -1) return;
```
https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L133

... traps focus on the GeoCoder input element and doesn't allow the user to tab through the following elements cause accessibility issues.

timchambers:
This seems to be the source of a problem I'm experiencing, too. If I have a map and geocoded inside a form and I hit return in one of the other form fields, the focus is stolen back to the geocoded and the submit event doesn't get fired. Any way to disable this focus stealing?

andrewharvey:
Does anyone have an minimally reproducing example?

3zzy:
@andrewharvey Here: https://jsfiddle.net/iaezzy/zevaw734/8/

If you try navigating via keyboard using Tab key, once you reach the geocoder element you can't focus the next button or the previous (using Shift + Tab) because the focus is trapped.

jdaudier:
I have the same bug. Does anyone have a workaround for this while we are waiting for a fix?

3zzy:
@jdaudier Here's what I'm doing temporarily:

```
$(".mapboxgl-ctrl-geocoder input[type='text']").on('focus keydown', function (e) {  
    if (e.shiftKey && e.keyCode == 9) {
        // focus prev
    }
    else if (e.keyCode == 9) {
        focus next
    }
});
```

jdaudier:
@3zzy Thanks! It works for me but I didn't need to add the event on focus. Just curious about your reason for that.

scottsfarley93:
It looks like we could tackle this by changing the `suggestions` library with a relatively minor PR that prevents trapping `tab` events.

Specifically, changing https://github.com/tristen/suggestions/blob/gh-pages/src/suggestions.js#L59 and https://github.com/tristen/suggestions/blob/gh-pages/src/suggestions.js#L67 so that `keyCode==9` checks are no longer cases in the switch statements.

I'm not sure if this will cause any other unintended consequences though. I'll have to dig in a little more.

scottsfarley93:
This was done in #220 and will be fixed in version 4.

PR #220, opened 2019-03-19
[v4] fix focus trap

katydecorah:

Fixes focus trap described in #162. This PR makes sure that `_onKeyDown`, if the user is tabbing though do not `_clear` the input because it will set the focus back on the input.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - `trackProximity` turned on by default [#195](https://github.com/mapbox/mapbox-gl-geocoder/issues/195)
 - Bump suggestions to v1.3.4
 - Fix duplicate event bug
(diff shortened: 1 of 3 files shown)


katydecorah:
@scottsfarley93 I added a test to make sure that when the input is empty and in focus and then there is a keydown event from <kbd>Tab</kbd> that the `_clear` function is not called. The `_clear` function inside `_onKeyDown` sets the focus back on the input, which is why we are getting trapped on the input.

I attempted to test under the same conditions, expect that if a different key is pressed that `_clear` _is_ called. But I'm having trouble getting it to work because `_onKeyDown` is wrapped in `debounce()`. I'm not well versed in Sinon yet, but I read that we can use Sinon's timers. I'm struggling to figure it out 🙃. Do you have any suggestions on testing a debounce function or do you think the current test is sufficient for now?

Issue #89, opened 2016-12-01
mapbox-gl-geocoder mobile styles


vincentsarago:
The geocoder plugin may appear quite big especially on mobile or when you have already other buttons on the map controls 

Today style:
![geocoder](https://cloud.githubusercontent.com/assets/10407788/20799080/1cd2f422-b7af-11e6-9a15-1a140bf6a43b.jpg)

I think it will be good to have at least an `collapsed` option to make the geocoder plugin small or big like in http://www.liedman.net/leaflet-control-geocoder/

`wanted style` 
![capture d ecran 2016-12-01 a 10 03 04](https://cloud.githubusercontent.com/assets/10407788/20799230/8ebbbb14-b7af-11e6-973d-1b35e421ca0a.png)

tristen:
I'm inclined to require this from the implementor rather than building in an option to support this (see https://www.mapbox.com/imagery-requests/ as example)

vincentsarago:
all good! 👌 
can we maybe update the documentation ?

tristen:
Yeah ... I wonder how we should document this ...

yuletide:
I'm having issues with the control working on mobile at all. Often it won't respond to touch events, and the autocomplete results are hard to see/interact with. Adding some responsive styling or docs for how to add them later would be very helpful.

katydecorah:
Noting that we added a collapse option in the upcoming version 4, ref https://github.com/mapbox/mapbox-gl-geocoder/pull/222.

Leaving this issue until we cover mobile styling in version 4 as well.

scottsfarley93:
We addressed these issues in #233 and #222. Mobile styles should be better handled once version 4 . is released.

PR #222, opened 2019-03-19
[v4] create collapsed option

katydecorah:

This PR adds a `collapsed` option that will collapse the geocoder controller into a button. When hovered or focused, the controller will open up. It will collapse once the user focuses or hovers away from the controller, but only if they have cleared the search.

refs https://github.com/mapbox/mapbox-gl-geocoder/issues/89#issuecomment-375500355


API.md
```diff
@@ -28,6 +28,7 @@ A geocoder component using Mapbox Geocoding API
         properties. Search results closer to this point will be given
         higher priority.
     -   `options.trackProximity` **[Boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** If true, the geocoder proximity will automatically update based on the map view. (optional, default `true`)
+    -   `options.collapsed` **[Boolean](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Boolean)** If true, the geocoder control will collapse unless until hovered or in focus. (optional, default `false`)
```
andrewharvey: "unless until", maybe just "until"?

katydecorah: thank you!

test/test.geocoder.js
```diff
@@ -513,5 +513,54 @@ test('geocoder', function(tt) {
     t.end();
   });
 
+  tt.test('options.collapsed=true', function(t) {
```
scottsfarley93: nitpick: these tests look good, but should we move them to `test.ui.js` since they generally handle the testing of UI elements?

jayenashar:
the `collapsed` option doesn't seem to work when more than 640px.  Found this in the CSS overriding the width:

```
/* Media queries*/
@media screen and (min-width: 640px) {
  .mapboxgl-ctrl-geocoder {
    width: 33.3333%;
```

Also, if it's in a block of text, it may wrap to the next line when expanded, causing it to be unhovered, causing it to collapse, causing it to be hovered, causing it to expand, ...

Not sure how to take a video on my computer, but here's some screenshots.  I can't stay hovered over both versions.
collapsed
expanded

Issue #176, opened 2019-01-21
Set different options after initialization


hedderich:
Hey everyone,

is there any way or interface to update the placeholder text to a different string than the one provided when initializing a `MapboxGeocoder` object, without creating a new instance?

When reassigning `options.placeholder`, the value changes, however it is not reflected in the state of the text input. Applying the same logic to the language attribute works though.

Consider this example:

```javascript
const geocoder = new MapboxGeocoder({
  accessToken,
  language: 'en',
  placeholder: 'My placeholder when the app loads',
  ...
})

// user performs an action

geocoder.options.language = 'pt'
// language is set to Portuguese

geocoder.options.placeholder = 'My placeholder when the user performed an action'
// nothing happens
```

andrewharvey:
geocoder.options isn't exposed as part of the public API https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md so there's no guarantees on what happens if you change it directly.

The standard way to do this would be using setters/getters for each of those options, so geocoder.setPlaceholder('your placeholder'), but those haven't been implemented, only get/setProximity have been implemented.

I can see how this would make it easier to change language without needing to remove, and re-add a new geocoder.

tristen:
Yeah hmm, perhaps something like a `geocoder.setOptions({})` could work.

scottsfarley93:
This was addressed in #226. That work introduced get and set functions for most options set-able in the constructor. It will be in the version4 release.

Issue #197, opened 2019-03-13
Upcoming major v4 release


scottsfarley93:
There are several outstanding issues and feature improvements that would improve the usability, look/feel, and result relevance of the plugin. I suspect that some of these changes may be large and/or breaking, so I think targeting this as a v4 major release seems reasonable. 

Here's the list of tickets that we've prioritized but totally open to suggestions about other features/bugs that should also make the list. 

mapbox/mapbox-gl-geocoder#51 - camera animation properties
mapbox/mapbox-gl-geocoder#99 - result event emitted twice
mapbox/mapbox-gl-geocoder#115 - expose the clear method as public
mapbox/mapbox-gl-geocoder#162 -  accessibility issues
mapbox/mapbox-gl-geocoder#153 -  include matching text
mapbox/mapbox-gl-geocoder#176 - expose setOptions method
mapbox/mapbox-gl-geocoder#150 -  localize placeholder text 
mapbox/mapbox-gl-geocoder#89 - mobile styling
mapbox/mapbox-gl-geocoder#167 - minor style tweak
https://github.com/mapbox/mapbox-gl-geocoder/issues/157 - async local geocoder
#195 - new default options
#196 - feedback widget

\cc @mapbox/search @katydecorah @yuletide @kbauhaus @andrewharvey

scottsfarley93:
Noting several other features/fixes that will be included:
- #219 - show a mapbox-gl marker on the map
- #216 - update dev dependencies to avoid security vulnerabilities
- #217 - remove checks for unstable feature ids

andrewharvey:
To help track this, should we close issues when they are merged into the version4 branch?

scottsfarley93:
@andrewharvey we can certainly do that. It might be confusing, however, since the changes will not be deployed in the plugin until version 4 lands in master -- they'll just be waiting in the version 4 working branch. Happy to do whatever makes the most sense.

andrewharvey:
GL JS closes tickets once the PR is merged, not when the release is made. I think that's a good model, it makes it clearer which issues still need to be implemented and which are already done.

My vote would be we close them once merged into v4 branch.

scottsfarley93:
👍 we can start doing this.

scottsfarley93:
[...]

PR #226, opened 2019-03-20
Get & set options

scottsfarley93:
This PR fixes https://github.com/mapbox/mapbox-gl-geocoder/issues/176 by adding getter/setter methods.

This PRs get/set methods for the following options:
- `language`
- `zoom`
- `flyTo`
- `placeholder`
- `bbox`
- `countries`
- `types`
- `minLength`
- `limit`
- `filter`
- `proximity` (already implemented)

It's an open question as to whether we should support `get`, and particularly `set`, methods for these options, since changing their behavior more fundamentally alters the behavior of the plugin -- and I'm not sure of the use case of changing them after creating an instance with different values:

- `accessToken`
- `origin`
- `localGeocoder`
- `reverseMode`
- `reverseGeocode`

#208 will add get/set functions for html rendering functions and we should add analogous methods in #219 to support new options that we're adding in version 4 (#197)

\cc @katydecorah @yuletide


lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * Get the zoom level the map will move to when there is no bounding box on the selected result
+   * @returns {Number} the map zoom
+   */
+  getZoom: function(){
+    return this.options.zoom;
+  },
+
+  /**
+   * Set the zoom level 
+   * @param {Number} zoom On geocoded result what zoom level should the map animate to when a `bbox` isn't found in the response. If a `bbox` is found the map will fit to the `bbox`.
```
katydecorah: Small copy suggestion:

```suggestion
   * @param {Number} zoom The zoom level that the map should animate to when a `bbox` isn't found in the response. If a `bbox` is found the map will fit to the `bbox`.
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+  /**
+   * Get the parameters used to fly to the selected response, if any
+   * @returns {MapboxGeocoder} this
+   */
+  getFlyTo: function(){
+    return this.options.flyTo;
+  },
+
+  /**
+   * Sets the flyTo options
```
katydecorah: Small edit for consistency:

```suggestion
   * Set the flyTo options
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @returns {MapboxGeocoder} this
+   */
+  setPlaceholder: function(placeholder){
+    this.placeholder = (placeholder) ? placeholder : this._getPlaceholderText();
+    this._inputEl.placeholder = this.placeholder;
+    return this
+  },
+
+  /**
+   * Gets the bounding box used by the plugin
```
katydecorah: For consistency:
```suggestion
   * Get the bounding box used by the plugin
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @returns {Array<Number>} the bounding box, if any
+   */
+  getBbox: function(){
+    return this.options.bbox;
+  },
+
+  /**
+   * Sets the bounding box to limit search results to
```
katydecorah:
```suggestion
   * Set the bounding box to limit search results to
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+  /**
+   * Get a list of the types to limit search results to 
+   * @returns {String} a comma separated list of types to limit to 
+   */
+  getTypes: function(){
+    return this.options.types;    
+  },
+
+  /**
+   * Set the countries to limit search results to
```
katydecorah: Should this be types?

```suggestion
   * Set the types to limit search results to
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @param {String} countries a comma separated list of countries to limit to
```
katydecorah: Should this also be types?

```suggestion
   * @param {String} types a comma separated list of countries to limit to 
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @returns {MapboxGeocoder} this
+   */
+  setMinLength: function(minLength){
+    this.options.minLength = minLength;
+    if (this._typeahead)  this._typeahead.minLength = minLength;
+    return this;
+  },
+
+  /**
+   * Get the limit value used by the plugin
```
katydecorah:
```suggestion
   * Get the limit value for the number of results to display used by the plugin
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @param {String} countries a comma separated list of countries to limit to 
+   * @returns {MapboxGeocoder} this
+   */
+  setTypes: function(types){
+    this.options.types = types;
+    return this;
+  },
+
+  /**
+   * Get the minimum length used in the plugin
```
katydecorah: Maybe call out what the minimum length refers to?

```suggestion
   * Get the minimum number of characters typed to trigger results used in the plugin
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @returns {Number}
+   */
+  getMinLength: function(){
+    return this.options.minLength;
+  },
+
+  /**
+   * Set the minimum length value used by the plugin
```
katydecorah:
```suggestion
   * Set the minimum number of characters typed to trigger results used by the plugin
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+  /**
+   * Get the limit value used by the plugin
+   * @returns {Number} the limit value
+   */
+  getLimit: function(){
+    return this.options.limit;
+  },
+
+  /**
+   * Set the limit value used by the plugin
```
katydecorah:
```suggestion
   * Set the limit value for the number of results to display used by the plugin
```

lib/index.js
```diff
@@ -447,33 +469,185 @@ MapboxGeocoder.prototype = {
...
+   * @returns {MapboxGeocoder} 
+   */
+  setLimit: function(limit){
+    this.options.limit = limit;
+    if (this._typeahead) this._typeahead.options.limit = limit;
+    return this;
+  },
+
+  /**
+   * Get the filter function used by the plugin
```
katydecorah: I'm not exactly sure what `filter` does in suggestions, maybe we could add a little more to add context?

andrewharvey: It filters the results shows :wink: See https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-limit-region/

PR #234, opened 2019-03-22
[work in progress] feedback link

katydecorah:

This PR adds a search feedback link. This feature can be disabled by passing `feature: false` to the geocoder.

image

Appending the link to the results has been challenging because:
1. ul.suggestions is absolutely positionedh
2. ul.suggestions clears children as the user types
3. suggestions closes the results once the user moves focus from the input, which means when you try to click the feedback link the results close before you can click and the feedback link travels on the page and _then_ you can click it. 

We were able to overcome 1 + 2 by adding a wrapper to suggestions. However, I'm still having an issue with 3.

* I added a workaround to prevent the results from closing, but need to finish building this out or else find another solution. https://github.com/mapbox/mapbox-gl-geocoder/blob/e67d176a58dfe5a0e37c69869a545c181873dae2/lib/index.js#L132-L134 
* The code above is overriding this event listener in suggestions: https://github.com/tristen/suggestions/blob/e91dfbe3aaa33e7578769bdc099682fdfe9ac27e/src/suggestions.js#L37-L39 that function closes the results when the user focuses away from the input unless they are selecting an item https://github.com/tristen/suggestions/blob/e91dfbe3aaa33e7578769bdc099682fdfe9ac27e/src/suggestions.js#L93-L97

I'm not sure yet if the solution is in gl-geocoder or suggestions.


katydecorah:
closing this for now since it didn't make it to the initial v4 release and it's quite a bit behind master now

Issue #115, opened 2017-04-20
make clear method public


andrewharvey:
The `_clear` method is private, this can be useful for applications extending the geocoder. Is there any reason not to make it public? I'm happy to make the change and add the documentation in a PR but wanted to understand if there was any reasoning behind it.

scottsfarley93:
Hey @andrewharvey curious if you've had any recent thoughts on this issue. If not, I can take a stab at doing this and get it into our version 4.0.0 release, but wanted to check first to see if you wanted to PR this change.

andrewharvey:
I still think it's a good idea to make this method public. I haven't done any work on this, so all yours if you wanted to PR it.

PR #235, opened 2019-03-27
Public clear method

scottsfarley93:
This PR tackles #115 by exposing the `_clear` method as a new public method `clear` so that developers can further extend the plugin. It updates the documentation accordingly.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -393,9 +393,10 @@ MapboxGeocoder.prototype = {
 
     return request;
   },
+
   /**
-   * Invoked on geocoder clear button is clicked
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -393,8 +393,12 @@ MapboxGeocoder.prototype = {
 
     return request;
   },
-
-  _clear: function(ev) {
+  /**
+   * Invoked on geocoder clear button is clicked
```
andrewharvey: Since this isn't the event, rather a method you can call to clear the geocoder something simple like "Clear the input" should be sufficient.

ev is optional, actually only really used internally

Issue #258, opened 2019-04-15
relatedTarget returns null when clicking on suggestions list 


samanpwbb:
There's a small order of operations issue with how geocoder interacts with suggestions. Normally, blur events have a `relatedTarget` property that refers to the element that user targetted in order to trigger the blur.

The ideal

PR #259, opened 2019-04-15
ensure search box collapses on blur if clearOnBlur

### LINKED ISSUE #257: If clearOnBlur:true and collapsed:true, geocoder doesn't collapse on blur
The geocoder should collapse on blur if both `collapsed: true`  and `clearOnBlur: true`. Right now, blur leaves geocoder in a buggy state where it's uncollapsed, but, on hover, it will collapse.

samanpwbb:
Previously, the geocoder search input would not collapse on blur when `closeOnBlur` and `collapsed` options are set to true. Now, the geocoder will correctly collapse on blur.
 
Also refactors code so there is only one blur listener. Closes #257

There is a future improvement to make to this code, which I ticketed here: https://github.com/mapbox/mapbox-gl-geocoder/issues/258

### DIFF
--- a/API.md
+++ b/API.md
@@ -51,6 +51,7 @@
         -   [Parameters][47]
     -   [off][48]
         -   [Parameters][49]
+-   [relatedTarget][50]
 
 ## MapboxGeocoder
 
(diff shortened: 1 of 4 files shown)


samanpwbb:
@scottsfarley93 for review plz 🙏

PR #261, opened 2019-04-17
Update API.md

easherma:
fix formatting for 'on' event listeners, just a minor docs tweak

### DIFF
--- a/API.md
+++ b/API.md
@@ -331,7 +331,8 @@ Subscribe to events that happen within the plugin.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
The upstream bug is https://github.com/documentationjs/documentation/issues/676 prior PR's here #125 #166, the problem is they get replaced every time we re-generate the API docs.

andrewharvey:
I'll close this because the bug will keep appearing when the docs are re-generated, maybe we can try to manually fix this up when the docs are generated... or try to find another solution.

PR #262, opened 2019-04-17
Patch event query string

scottsfarley93:
Patches a small bug introduced in #212 [change](https://github.com/mapbox/mapbox-gl-geocoder/pull/212/files#diff-71dcd327d0ca067b490b22d677f81966R170) that causes the queryString to be mis-recorded. 

 - [n/a] write tests for all new functionality
 - [n/a] run `npm run docs` and commit changes to API.md
 - [] update CHANGELOG.md with changes under `master` heading before merging

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -171,10 +171,13 @@ MapboxEventManager.prototype = {
       keyboardLocale: this.locale
     }
 
+    // get the text in the search bar
     if (event === "search.select"){
-      payload.queryString = geocoder.inputString
+      payload.queryString = geocoder.inputString;
+    }else if (event != "search.select" && geocoder._inputEl){
+      payload.queryString = geocoder._inputEl;
     }else{
-      payload.queryString = (geocoder._inputEl) ? geocoder._inputEl.value : geocoder.inputString;
+      payload.queryString = geocoder.inputString;
     }
     return payload;
   },
--- a/lib/events.js
+++ b/lib/events.js
@@ -59,7 +59,6 @@ MapboxEventManager.prototype = {
     }
     this.lastSentIndex = resultIndex;
     this.lastSentInput = payload.queryString;
-    console.log(payload);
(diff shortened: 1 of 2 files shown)


lib/events.js
```diff
@@ -167,10 +167,16 @@ MapboxEventManager.prototype = {
...
       // routing: search.routing, //todo --> add to plugin
-      queryString: (geocoder._inputEl) ? geocoder._inputEl.value : geocoder.inputString,
       mapZoom: zoom,
       keyboardLocale: this.locale
     }
+
+    if (event === "search.select"){
+      payload.queryString = geocoder.inputString
+    }else{
+      payload.queryString = (geocoder._inputEl) ? geocoder._inputEl.value : geocoder.inputString;
```
samanpwbb: Standard js code style is to not use parenthesis here, they're unnecessary.

lib/events.js
```diff
@@ -167,10 +167,16 @@ MapboxEventManager.prototype = {
       proximity: proximity,
       limit: geocoder.options.limit,
       // routing: search.routing, //todo --> add to plugin
-      queryString: (geocoder._inputEl) ? geocoder._inputEl.value : geocoder.inputString,
       mapZoom: zoom,
       keyboardLocale: this.locale
     }
+
+    if (event === "search.select"){
+      payload.queryString = geocoder.inputString
```
samanpwbb: you're missing a semicolon on this line.

lib/events.js
```diff
@@ -167,10 +167,16 @@ MapboxEventManager.prototype = {
...
+    }else{
+      payload.queryString = (geocoder._inputEl) ? geocoder._inputEl.value : geocoder.inputString;
+    }
+    return payload;
```
samanpwbb: code-style wise, it might be more readable to use an if / else if / else pattern rather than an if statement with a ternary inside one of the cases.

PR #276, opened 2019-05-17
Changes to flyTo API docs

mzdraper:

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -19,7 +19,7 @@ var subtag = require('subtag');
  * @param {String} options.origin Use to set a custom API origin. Defaults to https://api.mapbox.com.
  * @param {Object} [options.mapboxgl] A [mapbox-gl](https://github.com/mapbox/mapbox-gl-js) instance to use when creating [Markers](https://docs.mapbox.com/mapbox-gl-js/api/#marker). Required if `options.marker` is true.
  * @param {Number} [options.zoom=16] On geocoded result what zoom level should the map animate to when a `bbox` isn't found in the response. If a `bbox` is found the map will fit to the `bbox`.
- * @param {Boolean|Object} [options.flyTo] If false, animating the map to a selected result is disabled. If true, animating the map will use the default animation parameters. If an object, it should describe the destination and animation of the transition for flyTo. See [GL JS documentation](https://docs.mapbox.com/mapbox-gl-js/api/#map#flyto) for details.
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -19,7 +19,7 @@ var subtag = require('subtag');
  * @param {String} options.origin Use to set a custom API origin. Defaults to https://api.mapbox.com.
  * @param {Object} [options.mapboxgl] A [mapbox-gl](https://github.com/mapbox/mapbox-gl-js) instance to use when creating [Markers](https://docs.mapbox.com/mapbox-gl-js/api/#marker). Required if `options.marker` is true.
  * @param {Number} [options.zoom=16] On geocoded result what zoom level should the map animate to when a `bbox` isn't found in the response. If a `bbox` is found the map will fit to the `bbox`.
- * @param {Boolean|Object} [options.flyTo] If false, animating the map to a selected result is disabled. If true, animating the map will use the default animation parameters. If an object, the object will be passed to the map method to specify a custom animation when a result is selected.
+ * @param {Boolean|Object} [options.flyTo] If false, animating the map to a selected result is disabled. If true, animating the map will use the default animation parameters. If an object, it should describe the destination and animation of the transition for flyTo. See [GL JS documentation](https://docs.mapbox.com/mapbox-gl-js/api/#map#flyto) for details.
```
andrewharvey: >it should describe the destination and animation of the transition for flyTo. See [GL JS documentation](https://docs.mapbox.com/mapbox-gl-js/api/#map#flyto) for details.

Just a minor point, that if flyTo is an object, then you don't and can't control the destination through this option. In fact mapbox-gl-draw will either use `map.fitBounds` and pass it's own bounding box or `map.flyTo` and overwrite any `center` you set in the `flyTo` options.

It might be clearer to say specifically the options are passed to `fitBounds` and `flyTo`, where you can use any `AnimationOptions`, `CameraOptions` in addition to the fields supported by `fitBounds` and `flyTo`. I know this is a longer way to say this, but I think it's better to be clearer in the API documentation.

lib/index.js
```diff
@@ -670,15 +670,15 @@ MapboxGeocoder.prototype = {
 
   /**
    * Get the parameters used to fly to the selected response, if any
-   * @returns {MapboxGeocoder} this
+   * @returns {Boolean|Object} The flyTo option. If an object, it should describe the destination and animation of the transition for flyTo. See [GL JS documentation](https://docs.mapbox.com/mapbox-gl-js/api/#map#flyto) for details.
```
andrewharvey: Good catch

andrewharvey:
Thanks for your contribution @mzdraper, in an effort to get this PR merged, I've rebased from master again and made some tweaks to try and be even more specific.

PR #294, opened 2019-09-17
Update CONTRIBUTING.md with easier examples version script

andrewharvey:
fixes https://github.com/mapbox/mapbox-gl-js/pull/8760

The current links to the examples to update upon each release is getting longer, prone to having broken links.

Switching to an easier one liner which will update all GL JS examples using the geocoder.

PS I've tested this on Linux, I'm not sure if OSX has a different `sed`.


scottsfarley93:
👍 looks good to me. I tested on my OSX and I have `sed` installed.

Issue #270, opened 2019-05-06
Allow use of geocoder without map


yuletide:
We've had a few requests for using `gl-geocoder` without any map component, which is not currently possible.

antony:
This would be greatly helpful. for instance, in a search form, or on mobile, where there isn't always room for a map.

jakobeissler:
This would be awesome!!

scottsfarley93:
I started sketching out a solution to this over in #295. Would love feedback if anyone has thoughts on that approach.

scottsfarley93:
This was added in #295 and will be included in the upcoming v4.5.0 release.

PR #295, opened 2019-09-19
No Map

scottsfarley93:
This PR allows using the mapbox-gl-geocoder plugin without a map, as requested in https://github.com/mapbox/mapbox-gl-geocoder/issues/270. 

Currently, the plugin is added to the map using the map's `addControl` method. In order to facilitate adding the map to other elements, I added a new `addTo` method on the geocoder class. If the addTo method is called with a mapbox-gl-js map, it will be added like normal (the map's addControl method will be called under the hood). If, on the other hand, `geocoder.addTo` is called with a string that looks like a reference to an HTML class or ID, it will add the search bar to that element. In these cases, the geocoder will have no `_map` property and all map-related methods (like updateProximity or flyTo result) will not be executed. 

This still needs tests and I want to make sure there's no edge cases where the lack of a map causes unexpected problems. 

I will also need to:
- return the debug page to it's original state
- Add a new mapbox-gl-js example that shows how to use the geocoder without the map 

\cc @kbauhausness @shabnomnom 

### Launch Checklist

### DIFF
--- a/debug/index.js
+++ b/debug/index.js
@@ -18,12 +18,12 @@ insertCss(
 
 var MapboxGeocoder = require('../');
 
-// var mapDiv = document.body.appendChild(document.createElement('div'));
-// mapDiv.style.position = 'absolute';
-// mapDiv.style.top = 0;
-// mapDiv.style.right = 0;
-// mapDiv.style.left = 0;
-// mapDiv.style.bottom = 0;
+var mapDiv = document.body.appendChild(document.createElement('div'));
+mapDiv.style.position = 'absolute';
+mapDiv.style.top = 0;
+mapDiv.style.right = 0;
+mapDiv.style.left = 0;
+mapDiv.style.bottom = 0;
 
 var notMapDiv = document.body.appendChild(document.createElement('div'));
 notMapDiv.style.position = 'absolute';
@@ -35,12 +35,12 @@ notMapDiv.style.backgroundColor = 'red';
 
 notMapDiv.classList.add("notAMap");
 
-// var map = new mapboxgl.Map({
-//   container: mapDiv,
-//   style: 'mapbox://styles/mapbox/streets-v9',
-//   center: [-79.4512, 43.6568],
-//   zoom: 13
-// });
+var map = new mapboxgl.Map({
+  container: mapDiv,
+  style: 'mapbox://styles/mapbox/streets-v9',
+  center: [-79.4512, 43.6568],
+  zoom: 13
+});
 
 var coordinatesGeocoder = function(query) {
   var matches = query.match(/^[ ]*(-?\d+\.?\d*)[, ]+(-?\d+\.?\d*)[ ]*$/);
@@ -97,27 +97,27 @@ window.geocoder = geocoder;
 var button = document.createElement('button');
 button.textContent = 'click me';
 
-// var removeBtn = document.body.appendChild(document.createElement('button'));
-// removeBtn.style.position = 'absolute';
-// removeBtn.style.zIndex = 10;
-// removeBtn.style.top = '10px';
-// removeBtn.style.left = '10px';
-// removeBtn.textContent = 'Remove geocoder control';
+var removeBtn = document.body.appendChild(document.createElement('button'));
+removeBtn.style.position = 'absolute';
+removeBtn.style.zIndex = 10;
+removeBtn.style.top = '10px';
+removeBtn.style.left = '10px';
+removeBtn.textContent = 'Remove geocoder control';
 
-// map
-//   .getContainer()
-//   .querySelector('.mapboxgl-ctrl-bottom-left')
-//   .appendChild(button);
-// // map.addControl(geocoder);
+map
+  .getContainer()
+  .querySelector('.mapboxgl-ctrl-bottom-left')
+  .appendChild(button);
+// map.addControl(geocoder);
 
-// map.on('load', function() {
-//   button.addEventListener('click', function() {
-//     geocoder.query('Montreal Quebec');
-//   });
-//   removeBtn.addEventListener('click', function() {
-//     map.removeControl(geocoder);
-//   });
-// });
+map.on('load', function() {
+  button.addEventListener('click', function() {
+    geocoder.query('Montreal Quebec');
+  });
+  removeBtn.addEventListener('click', function() {
+    map.removeControl(geocoder);
+  });
+});
 
 geocoder.on('results', function(e) {
   console.log('results: ', e.features);
--- a/debug/index.js
+++ b/debug/index.js
@@ -25,16 +25,6 @@ mapDiv.style.right = 0;
 mapDiv.style.left = 0;
 mapDiv.style.bottom = 0;
 
-var notMapDiv = document.body.appendChild(document.createElement('div'));
-notMapDiv.style.position = 'absolute';
-notMapDiv.style.top = 0;
-notMapDiv.style.right = 0;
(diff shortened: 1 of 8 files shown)


debug/index.js
```diff
@@ -18,19 +18,29 @@ insertCss(
...
-mapDiv.style.left = 0;
-mapDiv.style.bottom = 0;
-
-var map = new mapboxgl.Map({
-  container: mapDiv,
-  style: 'mapbox://styles/mapbox/streets-v9',
-  center: [-79.4512, 43.6568],
-  zoom: 13
-});
+// var mapDiv = document.body.appendChild(document.createElement('div'));
```
andrewharvey: What about creating a different debug page such as `debug/mapless.js` so both can be tested side by side?

scottsfarley93: Yeah great idea, I'll add that to this PR.

scottsfarley93:
Tapping @yuletide for review.

andrewharvey:
Is it worth linking to the Mapbox TOS obligations when using places? [https://www.mapbox.com/legal/tos#[GAGA]](https://www.mapbox.com/legal/tos#[GAGA]), since you wouldn't be able to use no map unless you exclude places.

lib/events.js
```diff
@@ -85,13 +85,16 @@ MapboxEventManager.prototype = {
...
+
     //pass invalid event
     if (!keyEvent.key) return;
     // don't send events for keys that don't change the input
     // TAB, ESC, LEFT, RIGHT, ENTER, UP, DOWN
     if (keyEvent.metaKey || [9, 27, 37, 39, 13, 38, 40].indexOf(keyEvent.keyCode) !== -1) return;
     var payload = this.getEventPayload('search.keystroke', geocoder);
     payload.lastAction = keyEvent.key;
+    console.log("action: ", payload.lastAction)
+    console.log(payload)
```
yuletide:
```suggestion
```

lib/index.js
```diff
@@ -90,8 +90,43 @@ MapboxGeocoder.prototype = {
...
+      container.addControl(this)
+    }
+    // if the container is not a map, but an html element, then add the control to that element
+    else if (typeof container == 'string' && (container.startsWith('#') || container.startsWith('.'))){
+      const parent = document.querySelectorAll(container);
+      if (parent.length == 0){
+        throw new Error("Element ", container, "not found.")
+      }
+
+      parent.forEach(function(parentEl){
```
yuletide: I was able to create two geocodes with the same class applied to two different elements, but they behave unpredictably once created 
gif

lib/index.js
```diff
@@ -182,20 +217,20 @@ MapboxGeocoder.prototype = {
...
     this.mapMarker = null;
     this._handleMarker = this._handleMarker.bind(this);
-
-    this._mapboxgl = this.options.mapboxgl;
-    if (!this._mapboxgl && this.options.marker) {
-      console.error("No mapboxgl detected in options. Map markers are disabled. Please set options.mapboxgl.");
-      this.options.marker = false;
+    if (this._map){
+      this._mapboxgl = this.options.mapboxgl;
+      if (!this._mapboxgl && this.options.marker) {
```
yuletide:
```suggestion
      if (!this._mapboxgl && this.options.marker) {
       // eslint-disable-next-line no-console
```

lib/index.js
```diff
@@ -182,20 +217,20 @@ MapboxGeocoder.prototype = {
...
     this._typeahead.getItemValue = this.options.getItemValue;
 
-    if (this.options.trackProximity) {
+    if (this.options.trackProximity && this._map) {
       this._updateProximity();
       this._map.on('moveend', this._updateProximity);
     }
 
     this.mapMarker = null;
     this._handleMarker = this._handleMarker.bind(this);
```
yuletide: This would never be called if there's no map right? I'd move this stuff to the if (this._map) block below for simplicity. Also means you don't need to check for map in the handler

debug/nomap.js
```diff
@@ -0,0 +1,60 @@
...
+);
+
+var MapboxGeocoder = require('../');
+
+var notMapDiv = document.body.appendChild(document.createElement('div'));
+notMapDiv.style.position = 'absolute';
+notMapDiv.style.top = 0;
+notMapDiv.style.right = 0;
+notMapDiv.style.left = 0;
+notMapDiv.style.bottom = 0;
```
yuletide: To reproduce the double stuff add in:
```suggestion
notMapDiv.style.bottom = '50%';

var anotherNotMapDiv = document.body.appendChild(document.createElement('div'));
anotherNotMapDiv.style.position = 'absolute';
anotherNotMapDiv.style.top = '50%';
anotherNotMapDiv.style.right = 0;
anotherNotMapDiv.style.left = 0;
anotherNotMapDiv.style.bottom = 0;
anotherNotMapDiv.style.backgroundColor = 'green';
anotherNotMapDiv.classList.add('notAMap');
```

lib/index.js
```diff
@@ -90,8 +90,43 @@ MapboxGeocoder.prototype = {
...
+   * Add the geocoder to a container. The container can be either a mapboxgl.Map or a reference to an HTML class or ID. 
+   * 
+   * If the container is a mapboxgl.Map, this function will behave identically to Map.addControl(geocoder)
+   * If the container is an html id or class, the geocoder will be appended to that element
+   * 
+   * This function will throw an error if the container is not one a map or a class/id reference
+   * It will also throw an error if the referenced html element cannot be found in the document.body
+   * @param {String|mapboxgl.Map} container A reference to the container to which to add the geocoder 
+   */
+  addTo: function(container){
```
yuletide: Is this the new recommended way to invoke the geocoder? If so we should update the debug/index.js and docs to recommend using this method first, and not `onAdd` (which I think was always awkward so this is a welcome change)

lib/index.js
```diff
@@ -90,8 +90,43 @@ MapboxGeocoder.prototype = {
...
+    // if the container is a map, add the control like normal
+    if (container._controlContainer){
+      //  it's a mapbox-gl map, add like normal 
+      container.addControl(this)
+    }
+    // if the container is not a map, but an html element, then add the control to that element
+    else if (typeof container == 'string' && (container.startsWith('#') || container.startsWith('.'))){
```
yuletide: Suggestion: Allow passing in an HTMLElement

`if (container instanceof HTMLElement){parent = container}` and skip the `querySelector` bit

lib/index.js
scottsfarley93: I think the recommended method should still be `map.addControl(geocoder)` -- since I believe this will still be the primary usage. We can add an example that uses this method though.

debug/nomap.js
scottsfarley93: I added a test to `test.ui.js`, so I'm not sure we need to add this to the debug page.

lib/index.js
yuletide: Makes sense. So `map.addControl` if you have a map, otherwise use `addTo(container)`

This is the example I had in mind that we could update to use `addTo` instead of `onAdd` since its more logical: https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-outside-the-map/

pauloanalista:
Does anyone know if this issue has been resolved?

I want to replace Google Maps with MapBox and use a places autocomplete and I don't have a map.

scottsfarley93:
@pauloanalista this feature was released in v4.5.0. See some documentation [here](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md#addto).

Issue #293, opened 2019-09-16
result event not emitted for a two different results which share the same id


dhermyt:
Hi,

The steps to reproduce:

1.  Search for an address, like Bay Avenue 1. The map zooms in and `Result` event is called.
2.  Search for a nearby address, like Bay Avenue 2. The map recenters on the new location, but the `Result` is not called.

Why `Result` is not called in the second step? How to make the event raise on every search?

andrewharvey:
There was a bug in mapbox-gl-geocoder where the `result` event was fired twice for a single result, #99 and https://github.com/tristen/suggestions/issues/13. We could never work out how to properly fix that while still supporting IE11 and for all ways you can select a result.

So instead a fix was implemented that simply does a precheck to see if the result event being fired is for the same result, and if it is ignore it, a workaround for the underlying issue, but mostly works fine.

In this scenario those two addresses are returing the same `id`, because the `id` is the same mapbox-gl-geocoder things the second `result` event is just a duplicate of the first and hence does not fire.

The Geocoding API documentation at https://docs.mapbox.com/api/search/#geocoding-response-object is not entirely clear if this `id` in the response is absolutely unique or under which circumstances the same `id` is returned.

If this is simply an error in the upstream Mapbox Geocoding API, then we'll need to wait for a fix there.

If not and some results may have the same `id` we'll need to rethink how to address #99.

<details><summary>1 Bay Avenue</summary>
<p>
```
{
  "id": "address.3291212928152738",
  "type": "Feature",
  "place_type": [
    "address"
  ],
  "relevance": 1,
  "properties": {
    "accuracy": "interpolated"
  },
  "text_en-US": "Bay Avenue",
  "place_name_en-US": "1 Bay Avenue, Welland, Ontario L3B 3G3, Canada",
  "text": "Bay Avenue",
  "place_name": "1 Bay Avenue, Welland, Ontario L3B 3G3, Canada",
  "center": [
    -79.246267,
    42.949539
  ],
  "geometry": {
    "type": "Point",
    "coordinates": [
      -79.246267,
      42.949539
    ],
    "interpolated": true
  },
  "address": "1",
  "context": [
    {
      "id": "postcode.4926389830389270",
      "text_en-US": "L3B 3G3",
      "text": "L3B 3G3"
    },
    {
      "id": "place.3630735081406220",
      "wikidata": "Q424773",
      "text_en-US": "Welland",
      "language_en-US": "en",
      "text": "Welland",
[...]

Issue #281, opened 2019-06-14
multiple local geocoder search results broken


andrewharvey:
As reported at https://stackoverflow.com/questions/56579673/why-does-supplemental-mapbox-geocoder-allow-for-only-one-search in the example at https://docs.mapbox.com/mapbox-gl-js/example/forward-geocode-custom-data/ if you:

1. Search and select Lincoln Park
2. Search and select Burnham Park.

Upon 2, it doesn't actually move to Burnham park, which it should.

andrewharvey:
It looks like the issue could be around

https://github.com/mapbox/mapbox-gl-geocoder/blob/af44fb0f7458e709ec75ad7f8c1677f66f2ba84b/lib/index.js#L275

since localGeocoder sources may not have an id per feature set, we could either disable that where no id exists, or generate our own internal ids.

scottsfarley93:
This looks to be the same as https://github.com/mapbox/mapbox-gl-geocoder/issues/278. It seems that the `result` event is not being emitted on subsequent selections from the local geocoder. 

@andrewharvey good point about the potential lack of IDs from the localGeocoder. As you point out, I could see how the lack of IDs could cause no `result` event to be generated.

scottsfarley93:
This was fixed in #298 and will be included in the upcoming v4.5.0 release.

Issue #278, opened 2019-05-29
Local geocoder doesn't emit result event


scottsfarley93:
In some cases, the localGeocoder will not emit a `result` event, which can cause a bad user experience, since the map will not zoom to the selected result. 

Steps to replicate:
1. Fire up this example: https://docs.mapbox.com/mapbox-gl-js/example/forward-geocode-custom-data/
2. Search for "Park" 
3. Select one of the results
4.  Search for "Park" again
5. Select a new result --  notice that nothing happens

We should ensure that results from the localGeocoder are triggering all the events they should be. 

Note: this bug is present in v4.3.0, and so pre-dates the localGeocoder changes made in #275 and scheduled for v4.4.0 (#277). 

\cc @lobenichou

scottsfarley93:
This was fixed in #298 and will be included in the upcoming v4.5.0 release.

dehghani-mehdi:
@scottsfarley93 when you will release v4.5.0?

PR #298, opened 2019-10-01
Remove feature.id checks for event deduplication

scottsfarley93:
Deduplicating events by feature id was introduced in https://github.com/mapbox/mapbox-gl-geocoder/pull/229. While this prevents duplicate `result` events (https://github.com/tristen/suggestions/issues/13), it introduces (at least) two undesirable side effects. 

1. When using a localGeocoder, the result event will only be emitted once, manifesting in poor user experience by not zooming to the location of the second selection. (Bug reports: https://github.com/mapbox/mapbox-gl-geocoder/issues/281, https://github.com/mapbox/mapbox-gl-geocoder/issues/278). 
2. If features share the same ID (as reported in https://github.com/mapbox/mapbox-gl-geocoder/issues/293), subsequent selections will not emit the result event. 

This PR aims to fix the above two issues.

My initial thought is to use the `place_name` instead of the feature's ID as a field to deduplicate on. This should work for case (2) above, because the two addresses will have different place_names. However, for the local geocoder, if some features share the same place_name, the same behavior will continue to occur. While I expect that most users will have unique place names in their data, there is no requirement for uniqueness, so perhaps we should not rely on this alone.

Alternate approaches include:
(1) adding additional checks to see if the results came from the local Geocoder
(2) attempting (again) to solve the duplicate result emission in the suggestions library.  

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Master
+
+### Bug fixes 🐛
+
+- Fix event deduplication [#298](https://github.com/mapbox/mapbox-gl-geocoder/pull/298).
+
+
 ## v4.4.2
 
 ### Features / Improvements 🚀
--- a/lib/index.js
+++ b/lib/index.js
@@ -272,7 +272,7 @@ MapboxGeocoder.prototype = {
   },
   _onChange: function() {
     var selected = this._typeahead.selected;
-    if (selected  && selected.id !== this.lastSelected) {
+    if (selected  && selected.place_name !== this.lastSelected) {
       this._clearEl.style.display = 'none';
       if (this.options.flyTo) {
         var flyOptions;
@@ -311,7 +311,7 @@ MapboxGeocoder.prototype = {
 
       this._eventEmitter.emit('result', { result: selected });
       this.eventManager.select(selected, this);
-      this.lastSelected = selected.id;
+      this.lastSelected = selected.place_name;
     }
   },
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -308,10 +308,9 @@ MapboxGeocoder.prototype = {
       this._inputEl.focus();
       this._inputEl.scrollLeft = 0;
       this._inputEl.setSelectionRange(0, 0);
-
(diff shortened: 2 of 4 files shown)


scottsfarley93:
As per https://github.com/mapbox/mapbox-gl-geocoder/issues/293#issuecomment-534896938, it might make sense to use a stringified version of the whole feature as our de-duplication criterion. This would mean that users would not need to ensure uniqueness across their `place_name`s, because the coordinates and other feature data would make each feature unique. 

Tests to come.

andrewharvey:
I think this is a reasonable approach.

sashako:
Hello, hope it's okay that I butt in. I've encountered this bug in a bit of a different context: If the user decided to legitimately enter the same address twice the second search flow will not trigger 'result' event. Have you considered clearing out `this.lastSelected` on user input or perhaps with just a long enough timeout - so you're still deduping consecutive events that come in rapid succession, but enable the user to enter the same address twice if they choose to do so.

scottsfarley93:
Hey @sashako thanks for jumping in! Could you describe a bit more about this use case? In what cases is the user attempting to enter the same address twice in a row? I've not yet considered this.

I'm not opposed to resetting the `lastSelected` property on input change. I think that will still avoid the duplicate result events emitted by suggestions.

scottsfarley93:
@sashako I added a change in https://github.com/mapbox/mapbox-gl-geocoder/pull/298/commits/82be5ebc05f90df018f462e83ec635abb5094c41 to reset the lastSelected object on each keyDown event. Would you mind giving this a test and seeing if it fixes the issue you describe [here](https://github.com/mapbox/mapbox-gl-geocoder/pull/298#issuecomment-538483191), where you're unable to select the same feature twice in a row?

andrewharvey:
> @sashako I added a change in 82be5eb to reset the lastSelected object on each keyDown event. Would you mind giving this a test and seeing if it fixes the issue you describe here, where you're unable to select the same feature twice in a row?

I can't speak for @sashako, but I tested a use case where there is no `id` supplied, but the user can select the same result twice in a row, and this PR fixes the issue for me.

PR #300, opened 2019-10-02
Paste event

scottsfarley93:
Fixes the paste issue reported in #140 by adding an `onPaste` event to the geocoder. 

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -229,7 +229,6 @@ MapboxGeocoder.prototype = {
   _onPaste: function(e){
     var value = (e.clipboardData || window.clipboardData).getData('text');
     if (value.length >= this.options.minLength) {
-      console.log(value)
       this._geocode(value);
     }
   },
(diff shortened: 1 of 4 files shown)


lib/index.js
```diff
@@ -224,6 +226,13 @@ MapboxGeocoder.prototype = {
     return this;
   },
 
+  _onPaste: function(e){
+    var value = (e.clipboardData || e.clipboardData).getData('text');
```
yapalexei: This seems redundant `(e.clipboardData || e.clipboardData)`. Did you mean `(event.clipboardData || window.clipboardData)`? I believe `window.clipboardData` is the IE support though not considered safe/secure.

scottsfarley93: Good catch, I think that is supposed to be looking for `window.clipboardData`, I'll update the PR.

scottsfarley93:
@yapalexei would you mind taking this for a spin to see if it fixes the problem you described a workaround for [here](https://github.com/mapbox/mapbox-gl-geocoder/issues/140#issuecomment-537781532)?

yapalexei:
Sure. will do later this evening.

yapalexei:
@scottsfarley93 Checked it out, works great!

PR #301, opened 2019-10-14
only add a marker if center present in result

andrewharvey:

If the geocoding response only contains a `bbox` and no `center` then don't try to add a marker to the map (which would fail anyway).

I'd love to add a test, but because the tests currently query the live API, it would require a bit of work to put in place a mock instead.

 - [ ] ~~run `npm run docs` and commit changes to API.md~~

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -922,10 +922,12 @@ MapboxGeocoder.prototype = {
       color: '#4668F2'
     }
     var markerOptions = extend({}, defaultMarkerOptions, this.options.marker)
-    this.mapMarker = new this._mapboxgl.Marker(markerOptions);
-    this.mapMarker
-      .setLngLat(selected.center)
-      .addTo(this._map);
+    if (selected.center) {
+      this.mapMarker = new this._mapboxgl.Marker(markerOptions);
+      this.mapMarker
+        .setLngLat(selected.center)
+        .addTo(this._map);
+    }
     return this;
   },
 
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -915,10 +915,12 @@ MapboxGeocoder.prototype = {
       color: '#4668F2'
     }
     var markerOptions = extend({}, defaultMarkerOptions, this.options.marker)
-    this.mapMarker = new this._mapboxgl.Marker(markerOptions);
-    this.mapMarker
-      .setLngLat(selected.center)
-      .addTo(this._map);
+    if (selected.center) {
```
scottsfarley93: This looks good. 

We could also try to compute the center from the `geometry.coordinates` if the center property doesn't exist. We should always have coordinates, as the results are expected to be geojson.

andrewharvey: Yes but since I don't think it can be guaranteed to always be a point geometry, eg. searching for a street might return the whole street linestring, or for a country might return the country polygon, so it can be a bit more complicated. That would be a future enhancement but for now I think this simple check should suffice.

andrewharvey: Okay I've added a extra commit which will check for a Point geometry and use that if exists.

andrewharvey:
As an aside, I'd prefer to add CHANGELOG entries to the PR description rather than committing and then the person doing the release can collate them together into the CHANGELOG, because in this case GitHub thought it could cleanly rebase when merging, but that would have incorrectly put the CHANGELOG description in the wrong spot.

PR #302, opened 2019-10-14
allow results with no properties but with bbox

andrewharvey:

If a geocoding response lacks `properties` but has `bbox`, in master, it won't use the bbox. This PR fixes that to use a bbox over center when no properties exists.

 - [ ] ~~run `npm run docs` and commit changes to API.md~~

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -326,13 +326,7 @@ MapboxGeocoder.prototype = {
       this._clearEl.style.display = 'none';
       if (this.options.flyTo) {
         var flyOptions;
-        if (selected.properties && !exceptions[selected.properties.short_code] && selected.bbox) {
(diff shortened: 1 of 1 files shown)


andrewharvey:
I can update the changelog before merging but no point doing now as it'll just gather up merge conflicts.

andrewharvey:
> One question -- Since properties is a required member of valid geojson, is the idea behind this change to support localGeocoder results that may be in a format other than geojson?

Yes that's the idea, although `properties` is required, I think handling it's absence gracefully is better, especially when common use cases of the `localGeocoder` might return responses lacking a properties field.

andrewharvey:
I'll add a unit test next.

PR #304, opened 2019-10-16
Add example using the addTo method

adrianababakanian:
This PR adds an example to API.md for using the geocoder plugin without a map. The example clarifies that the `class`/`id` name must be appropriately prepended with `.` or `#` to be passed to the `geocoder.addTo` method.

### DIFF
--- a/API.md
+++ b/API.md
@@ -113,13 +113,20 @@ Returns **[MapboxGeocoder][68]** `this`
 
 ### addTo
 
-Add the geocoder to a container. The container can be either a mapboxgl.Map or a reference to an HTML class or ID. 
+Add the geocoder to a container. The container can be either a `mapboxgl.Map` or a reference to an HTML `class` or `id`. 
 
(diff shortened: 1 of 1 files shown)


scottsfarley93:
Thanks for the PR @adrianababakanian! 

The api.md file is autogenerated from function docstrings when you run the `npm run docs` command. I suspect that if you modify API.md directly, those changes will be overwritten the next time someone runs the doc generation command. The good news is that I think you can add the changes that you've put here into the comment string for the addTo method (https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L94-L100) and then run `npm run docs` and your comment will be propagated through, so the end result will be the same.

adrianababakanian:
Thank you for noting this @scottsfarley93 ! Closing this PR in favor of https://github.com/mapbox/mapbox-gl-geocoder/pull/305

PR #305, opened 2019-10-16
Add example using the addTo method

adrianababakanian:
This PR adds an example to API.md (by updating the function docstring in `lib/index.js` and running `npm run docs`) for using the geocoder plugin without a map. The example clarifies that the `class`/`id` name must be appropriately prepended with `.` or `#` to be passed to the `geocoder.addTo` method.


lib/index.js
```diff
@@ -91,19 +91,26 @@ MapboxGeocoder.prototype = {
...
+   * If the container is an HTML `id` or `class`, the geocoder will be appended to that element.
+   *
+   * This function will throw an error if the container is not either a map or a `class`/`id` reference.
+   * It will also throw an error if the referenced HTML element cannot be found in the `document.body`.
+   *
+   * For example, if the HTML body contains the element `<div id='geocoder-container'></div>`, the following script will append the geocoder to `#geocoder-container`:
+   *
+   * ```javascript
+   * var geocoder = new MapboxGeocoder({ accessToken: mapboxgl.accessToken });
+   * map.addTo('#geocoder-container');
```
scottsfarley93: I think this should be `geocoder.addTo(...`

adrianababakanian: My bad, you're correct! Will fix.

PR #306, opened 2019-10-17
addTo returns itself

ivalkenburg:
Tiny addition so addTo returns itself.


andrewharvey:
The API.md is automatically generated, I've updated the JSDoc with the documentation change.

andrewharvey:
I'm neutral on this change, so I'll defer to @scottsfarley93 who added this method for review.

PR #316, opened 2019-11-22
use placeholder value for input element aria-label

sethvincent:
Hello, this adds a small accessibility improvement: adding an aria-label to the input element.

I debated whether a new option would be best and ultimately reused the placeholder value as the aria-label value.

In most cases, and certainly in the default case of `Search`, it makes sense for the placeholder and the aria-label to be the same.

Happy to make any needed changes to get this merged.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -769,6 +769,7 @@ MapboxGeocoder.prototype = {
   setPlaceholder: function(placeholder){
     this.placeholder = (placeholder) ? placeholder : this._getPlaceholderText();
     this._inputEl.placeholder = this.placeholder;
+    this._inputEl.setAttribute('aria-label', this.placeholder);
     return this
   },
 


andrewharvey:
Thanks for the PR.

According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is not visible on the screen. If there is visible text labelling the element, use aria-labelledby instead.

So would that mean that when the placeholder text is present it should use aria-labelledby and once you start entering text and the placeholder dissapears than switch to using aria-label?

Otherwise would screen readers double announce the "search".

sethvincent:
From what I've read so far it seems that screen reader support for the placeholder attribute isn't guaranteed and in some cases isn't read at the right moment to be useful as a label.

The accessibility testing tools that I've tried don't recognize placeholder in their tests as a valid label.

> Use it in cases where a text label is not visible on the screen.

I'm pretty sure that when they say `text label` they mean the `<label>` element and not a placeholder attribute.

Here's one of the things I've been looking at as I've been learning about this: https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text

PR #319, opened 2019-11-28
Update API.md

CodeZeno:
Updated descriptions for `options.types` and `options.trackProximity`.

`options.types` had misleading information that was unnecessary.
`options.trackProximity` only works for zoom levels 10 and above which should be noted.


andrewharvey:
Thanks, the API.md is auto-generated by `npm run docs`, if you could made these changes in the source JS doc, and then re-gen the docs, and commit that?

andrewharvey:
superseded by #320

PR #320, opened 2019-11-28
Update API.md trackProximity and reverseGeocode/types methods

### LINKED ISSUE #317: API.md parameters options.type misleading and incorrect.
The [API.md](https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md#parameters) needs updating.
Under `options.types` it states:

> If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used.

This is misleading and incorrect. If you specify only one type and there are no results for that type you will get the message "No results found". You can and should specify multiple types in order of preference.

CodeZeno:
Updated descriptions for `options.types` and `options.trackProximity`.

`options.types` had misleading information that was unnecessary.
`options.trackProximity` only works for zoom levels 10 and above which should be noted.

closes #317

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -25,7 +25,7 @@ var subtag = require('subtag');
  * a geographical point given as an object with latitude and longitude
  * properties. Search results closer to this point will be given
  * higher priority.
- * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself. (optional, default `true`)
+ * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself.
  * @param {Boolean} [options.collapsed=false] If true, the geocoder control will collapse until hovered or in focus.
  * @param {Boolean} [options.clearAndBlurOnEsc=false] If true, the geocoder control will clear it's contents and blur when user presses the escape key.
  * @param {Boolean} [options.clearOnBlur=false] If true, the geocoder control will clear its value when the input blurs.
--- a/lib/index.js
+++ b/lib/index.js
@@ -25,7 +25,7 @@ var subtag = require('subtag');
  * a geographical point given as an object with latitude and longitude
  * properties. Search results closer to this point will be given
  * higher priority.
- * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view.
+ * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself. (optional, default `true`)
  * @param {Boolean} [options.collapsed=false] If true, the geocoder control will collapse until hovered or in focus.
  * @param {Boolean} [options.clearAndBlurOnEsc=false] If true, the geocoder control will clear it's contents and blur when user presses the escape key.
  * @param {Boolean} [options.clearOnBlur=false] If true, the geocoder control will clear its value when the input blurs.
@@ -37,7 +37,6 @@ var subtag = require('subtag');
  * @param {string} [options.types] a comma seperated list of types that filter
  * results to match those specified. See https://docs.mapbox.com/api/search/#data-types
  * for available types.
- * If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used.
  * @param {Number} [options.minLength=2] Minimum number of characters to enter before results are shown.
  * @param {Number} [options.limit=5] Maximum number of results to show.
  * @param {string} [options.language] Specify the language to use for response text and query result weighting. Options are IETF language tags comprised of a mandatory ISO 639-1 language code and optionally one or more IETF subtags for country or script. More than one value can also be specified, separated by commas.
--- a/API.md
+++ b/API.md
@@ -75,7 +75,7 @@ A geocoder component using Mapbox Geocoding API
         a geographical point given as an object with latitude and longitude
         properties. Search results closer to this point will be given
         higher priority.
-    -   `options.trackProximity` **[Boolean][62]** If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself. (optional, default `true`) (optional, default `true`)
+    -   `options.trackProximity` **[Boolean][62]** If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself. (optional, default `true`)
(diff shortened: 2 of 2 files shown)


lib/index.js
```diff
@@ -25,7 +25,7 @@ var subtag = require('subtag');
  * a geographical point given as an object with latitude and longitude
  * properties. Search results closer to this point will be given
  * higher priority.
- * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view.
+ * @param {Boolean} [options.trackProximity=true] If true, the geocoder proximity will automatically update based on the map view for zoom levels 10 and above. To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself.
```
andrewharvey: Sorry to be pedantic but technically this is any zoom > 9, which for example includes zoom `9.1`. I don't mind if we change the behaviour to be >= to make it clearer.

> "To enable proximity bias for all zoom levels you need to set the `options.proximity` yourself."
I'm hesitant to actually recommend this to users, since the Mapbox Geocoding API doesn't accept a zoom level in the proximity parameter when zoomed out at low zooms the proximity is more irrelevant.

What's the use case for passing a proximity even at low zooms?

CodeZeno: Ah yes true I guess it's "above 9"
Use case is we have a map that shows the 10 closest locations to the user and zoom out enough to show all 10 on screen. This might be a zoom level below 9.
We use IP based location so at times it can be inaccurate and pickup the ISP instead. As a user if I start typing my suburb I get results from other countries before results in the country that is on screen. I want locations closest to the center of the map still to take priority.
I feel like zoom level 9 is still very zoomed in and that something around 3+ would be more appropriate.

andrewharvey: The proximity used to only have an affect up to a few hundred meters, which is was the main driver behind only setting it >z9.

However I just checked carmen again now and it seems the logic has changed. https://github.com/mapbox/carmen/blob/0015a66dab6ba8a927cde06b744cb9991f15bd45/lib/util/proximity.js

Specifically addresses and pois are only affected by proximity weighted around z14, but country and region are weighted around z6.

So it looks like having a rough proximity even at low zooms is helpful for higher level results like localities, but I'm still concerned about an arbitrary map center having too much weight for low level results like addresses. Would be great if the Mapbox Geocoding Team could weigh in on this, or support zoom in the proximity bias.

CodeZeno: The map [here](https://docs.mapbox.com/search-playground/) works the way I would like.
If I first search "Australia" to center the map on the center of Australia.
I then search "Bris" I get "Brisbane, Queensland, Australia" or
If I search "107 Mullen" I get "107 Mullens Street, Balmain New South Wales 2041, Australia".
Both of these are exactly how I want it to work.
With mapbox-gl-geocoder I get results in other countries.

andrewharvey: Both of these are working as you prefer in the search playground precicly because the proximity isn't set. When you try at https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/ it's setting the proximity in Canada and so both of these searches aren't giving the same top results here.

ie. if you set the proximity at all zooms you'd be getting worse results when looking at a zoomed out map of the whole globe.

CodeZeno: If I customise the map to center on Australia like [this](https://codepen.io/codezeno/pen/LYYwoge?&editable=true) regardless of whether trackProximity is enabled or disabled or the zoom level. When I search "Bris" I get "Bristol, England, United Kingdom" as the first result. What is causing this difference compared to the playground map?

andrewharvey: it's the "language=en-US" which is causing the difference, mapbox-gl-geocoder sets it, the search playground doesn't (unless you turn it on). So it's not related to proximity.

CodeZeno: Ah you are absolutely right! This seems like flawed logic in the case of English.
How do I unset this? I have tried adding a IETF subtag but that didn't help and I cant find a way to unset it.

andrewharvey: Hmm seems you can't unset it, by default it uses "navigator.language || navigator.userLanguage || navigator.browserLanguage" https://github.com/mapbox/mapbox-gl-geocoder/blob/master/lib/index.js#L709

Let's cut this out to a new issue.

andrewharvey:
The discussion above is quite long so to summarise the commits so far in this PR

1. document what zoom levels trackProximity is enabled for. I'm :+1: to that.
2. direct users on how to workaround that zoom level conditional trackProximity. I think it's better to re-assess if we should still have a zoom level conditional proximity or not. If it turns out we shouldn't have it at all then no need to document how to remove it, and if it turns out we should have it we should document why.

@CodeZeno There is a 3rd change here relating to the reverseGeocode/types methods, I think it'll be easier to break that out into a separate PR as it's an unrelated change and it's too complicated to review both issues in a single PR.

> The proximity used to only have an affect up to a few hundred meters, which is was the main driver behind only setting it >z9.
> 
> However I just checked carmen again now and it seems the logic has changed. https://github.com/mapbox/carmen/blob/0015a66dab6ba8a927cde06b744cb9991f15bd45/lib/util/proximity.js
> 
> Specifically addresses and pois are only affected by proximity weighted around z14, but country and region are weighted around z6.
>
> So it looks like having a rough proximity even at low zooms is helpful for higher level results like localities, but I'm still concerned about an arbitrary map center having too much weight for low level results like addresses. Would be great if the Mapbox Geocoding Team could weigh in on this, or support zoom in the proximity bias.

scottsfarley93:
> I think it's better to re-assess if we should still have a zoom level conditional proximity or not. 

I suggest we continue conditional setting of proximity based on zoom level. The proximity parameter is designed primarily to facilitate local searches at a neighborhood-to-city scale. If we always set the proximity (even when the map is zoomed out), we'll end up with a case where we're strongly biasing towards a location (center of the map) that's not important to the user browsing the map at a low zoom. Conversely, if we don't set the proximity, we'll end up with a situation where users looking for features within their zoomed-in map view get results from all over the world. Conditionally setting the proximity at a medium-zoom seems like a good compromise and facilitates a good user experience at most zoom levels.

andrewharvey:
> I suggest we continue conditional setting of proximity based on zoom level. The proximity parameter is designed primarily to facilitate local searches at a neighborhood-to-city scale. If we always set the proximity (even when the map is zoomed out), we'll end up with a case where we're strongly biasing towards a location (center of the map) that's not important to the user browsing the map at a low zoom. Conversely, if we don't set the proximity, we'll end up with a situation where users looking for features within their zoomed-in map view get results from all over the world. Conditionally setting the proximity at a medium-zoom seems like a good compromise and facilitates a good user experience at most zoom levels.

Agreed that was the rational behind originally making the proximity zoom dependent. Of course the best solution is to pass the zoom together with the coordinates to the Geocoding API, though that's not yet supported.

With this in mind I still don't think it's a good idea to then tell people in the API docs to disable this unless there is a compelling reason to do so.

@CodeZeno I'd accept this if it was simply "If true, the geocoder proximity will automatically update based on the map view for zoom levels above 9." did you want to update this PR or should I make a new one?

And if you could split the other unrelated change out to another PR please?

andrewharvey:
I'll go ahead an close this PR since it looks like other recent changes have addressed some of the issues, feel free to open a new PR if there are still changes from here which should apply.

PR #327, opened 2019-12-14
disable proximity when reverseGeocoding

### LINKED ISSUE #318: turn trackProximity off by default when reverseGeocode is true
The Mapbox Geocoding API's Reverse Geocoder https://docs.mapbox.com/api/search/#reverse-geocoding doesn't support the proximity option, however when setting mapbox-gl-geocoder's options to `reverseGeocode: true`, it seems the `trackProximity` option is still `true` by default.

When `reverseGeocode` is set to true, `trackProximity` should be off by default.

andrewharvey:
closes #318. Since proximity is not supported by the reverseGeocoder, this disables it if it was set by trackProximity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## master
 
+### Bug fixes 🐛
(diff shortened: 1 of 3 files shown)


sawden:
This change causes errors when reverseGeocode is enabled.
image

PR #332, opened 2020-01-07
Fix formatting in list of available events for on listener

tnorthcutt:

Previously, the `clear` event was not on its own line. Now it is.

### DIFF
--- a/API.md
+++ b/API.md
@@ -374,7 +374,8 @@ Subscribe to events that happen within the plugin.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
The upstream bug is documentationjs/documentation#676 prior PR's here #261 #125 #166, the problem is they get replaced every time we re-generate the API docs. I don't think it's worth the effort manually patching this each time, so better to fix this either in documentation or the source file so that it get generated without this issue.

PR #343, opened 2020-03-13
Update mapbox-gl-geocoder.css

joewoodhouse:
Remove weird escape character which was causing my build (when importing this file into another SASS file) to fail.

### DIFF
--- a/lib/mapbox-gl-geocoder.css
+++ b/lib/mapbox-gl-geocoder.css
@@ -43,7 +43,7 @@
   color: rgba(0, 0, 0, 0.75);
   outline: 0;
   box-shadow: none;
-  outline: thin dotted\8;
+  outline: thin dotted;
 }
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
@katydecorah looks like this was something you introduced, I assume just an accident and we can safely remove it?

katydecorah:
This was introduced in https://github.com/mapbox/mapbox-gl-geocoder/commit/450a56734c3b6f6ddd776441a6738dd6e8b43eab - I believe it's a CSS hack to target certain versions of IE (ref: https://code.tutsplus.com/tutorials/quick-tip-how-to-target-ie6-ie7-and-ie8-uniquely-with-4-characters--net-10575). I think it's safe to remove, but would recommend that this branch is tested in IE to double check.

andrewharvey:
The build failure is unrelated so I'll merge.

PR #345, opened 2020-03-13
#344: Fix for results not firing on setInput

### LINKED ISSUE #344: Geocoder does not show results on setInput
When programatically setting the input with `setInput(searchInput)`, the search box is updated with the input value, however no suggestions are displayed.

It seems like `setInput(searchInput)` should be calling `this._geocode(searchInput)` instead of `this._onChange()`, since `this._onChange()` essentially does nothing as a result of the `this._typeahead.selected = null; `

https://github.com/mapbox/mapbox-gl-geocoder/blob/e2d22659a1493455fb16cd0d1a7b3bf76752e6dd/lib/index.js#L659-L666

GravlLift:
This change causes calls to `setInput` to behave in the same manner as a `keyDown` event, that is returning suggestions based on the input.

closes #344

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -661,7 +661,9 @@ MapboxGeocoder.prototype = {
     this._inputEl.value = searchInput;
     this._typeahead.selected = null;
     this._typeahead.clear();
-    this._geocode(searchInput);
+    if (searchInput.length >= this.options.minLength) {
+      this._geocode(searchInput);
+    }
     return this;
   },
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -661,7 +661,7 @@ MapboxGeocoder.prototype = {
     this._inputEl.value = searchInput;
     this._typeahead.selected = null;
     this._typeahead.clear();
-    this._onChange();
+    this._geocode(searchInput);
     return this;
   },
 


AbeCole:
Is there any method that exists to set the input and not query? My use case is geolocating a user manually and focusing on that location the same way I would do a selected result, but i'd like the input to show something like "--- User Location ---" (or their coords) so the UI is consistent and can be cleared the same way. I thought this would be possible with `geocoder.setInput('--- User location ---')`, however this now performs a search then returning & displaying the "No results" info.

alexandermorgan:
What if you passed a localGeocoder function that has a "--- User Location ---" result so that it will get suggested? Then in the function set the geometry.coordinates to the user's location. Or you could set the placeholder to "--- User Location ---" and set it back whenever you're done, but that's maybe a little roundabout.

AbeCole:
> What if you passed a localGeocoder function that has a "--- User Location ---" result so that it will get suggested? Then in the function set the geometry.coordinates to the user's location. Or you could set the placeholder to "--- User Location ---" and set it back whenever you're done, but that's maybe a little roundabout.

In the end I've gone with storing the 'user location active' in my state and overlaying a button to 'cancel' that, then applying the search to mapbox myself (using the same function the geocoder result goes to). I did this as it seemed like using a 'fake' location in the geocoder maybe the wrong approach at the moment.

zokioki:
What's the difference between `setInput`'s new behavior and using [`query`](https://github.com/mapbox/mapbox-gl-geocoder/blob/c15cb7b6fa07da0074b2604153c705dceabe5716/API.md#query)? My understanding was that you use `setInput` to set the new value and `query` to set the new value and fetch results for it.

andrewharvey:
> What's the difference between setInput's new behavior and using query? My understanding was that you use setInput to set the new value and query to set the new value and fetch results for it.

Good question, I don't know to be honest. If you wanted to review the source code you could. I know that when I checked now some of the private methods are lacking documentation so it can be hard to understand the intent.

M1ke:
This does feel like a breaking change even if not a bug - the old behaviour was stable for some time and documentation didn't suggest that `setInput` should open the list, so one shouldn't have expected it. Suddenly an innocent `npm update` is changing the behaviour in our application because of this. Likely it should have only been incorporated in a major version bump.

Is there a setting (or a way to add a setting?) that can revert to the old behaviour? I'd happily submit a PR if you'd accept a conditional value as a 2nd parameter to `setInput` that defines if it shows the list or not.

andrewharvey:
@M1ke more eyes on this repo would be great, and help reviewing PRs like this one appreciated. I've tried my best to keep this library maintained but am stretched thin at the moment. I think for this issue it's harder because it's about defining what should be the expected behaviour and needs to take a high level view. If you think you can improve something, please do open a PR and provide all the details on what the change is, why it should be made, and all the implications of the change. Just will depend who has the time to review it.

antony:
@andrewharvey I've gone ahead and reviewed the PR and it looks good, in case that's useful.

PR #350, opened 2020-04-18
Fix geocoder for maps that are inititalised on a lat or lng of 0

### LINKED ISSUE #253: Correctly format proximity parameter in url
The proximity parameter is not correctly included in the url when the longitude is zero.

I discovered this issue by setting the proximity parameter from `map.getCenter()`.

After looking at the code, this seems like an issue with the underlying geocoder service imported as: `var mbxGeocoder = require('@mapbox/mapbox-sdk/services/geocoding');`.

## Examples
`geocoder.setProximity([0, 2.0e-14]);` results in a request that only passes one number to the API—the one in scientific notation:
```
https://api.mapbox.com/geocoding/v5/mapbox.places/berli.json?limit=5&proximity=2e-14&language=en-US
```

Using a zero and a number without scientific notation behaves similarly `geocoder.setProximity([0, 0.2])`.

These cause the following server error:
image

## Expected behavior

The proximity parameter should contain both numbers. If scientific notation is a problem for our server, it should convert the number to a valid format.

## Possible Causes

*Falsy Values*
The leading zero value isn't included in the query parameters. Maybe we are incorrectly checking for the existence of a proximity parameter when generating the url and ignoring falsy values like zero.

sheridanvk:
Hi there! I ran into a bug when trying to use a geocoder object on a map initialised to London (lng: 0, lat: 51.5). I realised it was due to the way that zero values in lat/lngs were being handled in the proximity feature. 

This PR should fix that issue, but it also needs to be tested in conjunction with this change once it's merged: https://github.com/mapbox/mapbox-sdk-js/pull/362, as it turned out there was a similar bug in the SDK that also needed fixing before this would work correctly. 

Tests are included and can be made to pass locally if you clone my fork of @mapbox/mapbox-sdk-js (git@github.com:sheridanvk/mapbox-sdk-js.git) and take the following steps: 
in your `mapbox-sdk-js` folder, run:
`npm link`
in your `mapbox-gl-geocoder` folder, run:
`npm link @mapbox/mapbox-sdk-js`

Setting this to Draft status until the SDK change is merged - let me know if I should do anything differently there.

closes #253

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mapbox-gl-geocoder",
-  "version": "4.5.2",
+  "version": "4.5.1",
   "description": "A geocoder control for Mapbox GL JS",
   "main": "lib/index.js",
   "unpkg": "dist/mapbox-gl-geocoder.min.js",
--- a/package.json
+++ b/package.json
@@ -59,11 +59,11 @@
     "mapbox-gl": ">= 0.47.0 < 2.0.0"
   },
   "dependencies": {
-    "@mapbox/mapbox-sdk": "^0.5.0",
+    "@mapbox/mapbox-sdk": "^0.11.0",
     "lodash.debounce": "^4.0.6",
-    "suggestions": "^1.6.0",
-    "subtag": "^0.5.0",
     "nanoid": "^2.0.1",
+    "subtag": "^0.5.0",
+    "suggestions": "^1.6.0",
     "xtend": "^4.0.1"
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mapbox-gl-geocoder",
-  "version": "4.5.1",
+  "version": "4.5.2",
   "description": "A geocoder control for Mapbox GL JS",
   "main": "lib/index.js",
   "unpkg": "dist/mapbox-gl-geocoder.min.js",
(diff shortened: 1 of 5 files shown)


andrewharvey:
Looks good, should we also add a test for setting the proximity to 0?

sheridanvk:
Sounds good, added that test!

Should I leave this as draft for now until you roll the SDK fix out (thanks for approving that one!)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/mapbox-gl-geocoder",
-  "version": "4.5.1",
+  "version": "4.5.2",
```
andrewharvey, line 3: lets omit this this PR since it will be done as part of the release per https://github.com/mapbox/mapbox-gl-geocoder/blob/master/CONTRIBUTING.md#deploying

sheridanvk: aha makes sense, thanks. Will remove from here and package-lock.json. 

Does this line need to be removed too until the release is cut?
https://github.com/mapbox/mapbox-gl-geocoder/pull/350/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR3

andrewharvey: yes, you can leave it under the master header until the release is cut and then the entry can be made.

andrewharvey:
hi @sheridanvk thanks for your patience, mapbox-sdk-js now has v0.11.0 released which included your fix. If you wanted to update your PR here to upgrade the mapbox-sdk-js dependency then I can take another look and get this merged.

sheridanvk:
nice, thank you for getting it over the line!

PR #353, opened 2020-04-24
prepare v4.6.0

andrewharvey:
- [x] briefly describe the changes in this PR
 - ~~[ ]~~ write tests for all new functionality
 - [x] run `npm run docs` and commit changes to API.md
 - [x] update CHANGELOG.md with changes under `master` heading before merging

PR #356, opened 2020-04-28
remove MapboxAccessToken from travis global env since it's already set

andrewharvey:
as noted at https://github.com/mapbox/mapbox-gl-geocoder/pull/353#issuecomment-620328424, on Travis, the Mapbox token is already set globally and so we need to remove this to avoid it being re-set to '' and hence breaking builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
   - '6'
-env:
-  global: MapboxAccessToken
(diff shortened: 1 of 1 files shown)


andrewharvey:
@tristen good detective work on your end :detective: ! This looks like it does indeed work. The tests are still broken but that's a different issue and can be followed up in another PR.

Issue #379, opened 2020-07-01
fix build failure blocking release


andrewharvey:
```
# Geocoder--no map
# result was added to container
not ok - this is undefined
http://localhost:40937/script.js:448
      const el = this.onAdd(); //returns the input elements, which are then added to the requested html container
             ^
```

https://github.com/mapbox/mapbox-gl-geocoder/pull/353

andrewharvey:
due to a change introduced in #311. looks like function may need to be bound to this. It was before but that was dropped in the PR.

andrewharvey:
alternatively we can revert #311 which takes the pressure off fixing this issue see https://github.com/mapbox/mapbox-gl-geocoder/pull/311#issuecomment-657163138.

Myself I don't have the time to properly fix the issue but also would like to see the other bug fixes make a release so if there's no fix soon I might opt to revert those changes out of master back into a PR to unblock the release.

antony:
Ok, I've fixed/corrected/removed the various test failures + functionality.

PR #382, opened 2020-07-12
fix failing test which was missing bind, add tests for element scenario

antony:
This fixes https://github.com/mapbox/mapbox-gl-geocoder/issues/379 which was caused by my previous PR not binding the inner method correctly.

I'm still having failures with `.flyTo` locally but I get these on master anyway so I'm quite sure they're not related to my fix.

I am getting one failure about adding to a container (but it doesn't relate to my change), so just trying to determine if this is something to do with the PR or not.

Opening this for review whilst I do that work.

### DIFF
--- a/test/test.ui.js
+++ b/test/test.ui.js
@@ -401,7 +401,7 @@ test('Geocoder#inputControl', function(tt) {
   tt.end();
 });
 
-test('Geocoder--no map', function(tt) {
+test('Geocoder#addTo(String) -- no map', function(tt) {
   var container, geocoder;
 
   var changeEvent = document.createEvent('HTMLEvents');
@@ -430,7 +430,6 @@ test('Geocoder--no map', function(tt) {
     container.remove();
     t.end(); 
   });
-
 
   tt.test('input works without a map', function(t) {
     setup({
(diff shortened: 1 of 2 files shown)


antony:
Ok solved. None of my `flyTo` tests pass, but they don't work with the current master, or the previous release either, so it must just be my machine?

test/test.ui.js
```diff
@@ -515,26 +546,13 @@ test('Geocoder#addTo', function(tt) {
...
-    container.className = "notAMapEither"
     document.body.appendChild(container)
     geocoder = new MapboxGeocoder(opts);
-    geocoder.addTo(document.querySelectorAll(".notAMapEither"));
+    geocoder.addTo(container);
     t.ok(Object.keys(container.getElementsByClassName("mapboxgl-ctrl-geocoder--input")).length, 'geocoder exists when added to an html element')
     t.end(); 
   });
 
-  tt.test('throws if added to an unknown element', (t)=>{
```
andrewharvey, line 526: I agree, I'm not sure what this unit test is for. @scottsfarley93 added it in https://github.com/mapbox/mapbox-gl-geocoder/commit/7f53ecf331972161e12691e0fd9e94974d3ff46a#diff-23a1ea5628fe499863040c05a9d338aaR440 but as far as I can tell it's checking to make sure the element is a map, but in that PR at https://github.com/mapbox/mapbox-gl-geocoder/pull/295 support was added to allow adding the geocoder to a non-map element for a mapless geocoder.

andrewharvey:
Thanks @antony and yes don't worry about the flyTo failing test, that's unrelated and will need to be fixed another time.

PR #383, opened 2020-07-12
improve addTo documentation

andrewharvey:


andrewharvey:
@antony would you mind giving this a quick check to make sure this is okay per your changes?

antony:
Yep this looks great! Thanks :)

PR #165, opened 2018-08-31
Support async  using promises

yuletide:
This PR adds support for asynchronous `localgeocoder` functions using Promises. 

As it stands, this will likely require additional polyfills to prevent breakage in older browsers. Not sure the best approach for that but open to ideas before we merge.

closes #157

PR #385, opened 2020-07-30
Add option for a promise based externalGeocoder

mapbox-danny:
Proposal to add an option for an externalGeocoder, designed to accept a promise resolving to an array of features.  This Update would allow a server-side api/resource to be called asynchronously.


andrewharvey:
hi @mapbox-danny did you see https://github.com/mapbox/mapbox-gl-geocoder/pull/165#issuecomment-476455200 where we discussed leaving localGeocoder as is and creating an additional externalGeocoder which supports a promise. The idea being localGeocoder then works with local sources you already have loaded, and externalGeocoder where you have an external API to use.

lib/index.js
```diff
@@ -488,11 +482,29 @@ MapboxGeocoder.prototype = {
...
+            return localGeocoderRes.then(function(features) {
+              res.features = res.features ? features.concat(res.features) : features;
+              return res;
+            }, function(){
+              // on error, display the original result
+              return res;
+            });
+          } else {
+            // supplement Mapbox Geocoding API results with locally populated results
+            res.features = res.features ? localGeocoderRes.concat(res.features) : localGeocoderRes;
```
andrewharvey, line 501: I think where the localGeocoder returns a result faster than the Geocoding API (which it should if it's populated from a local source) then it should display the results immediately (like it currently should) instead of waiting for the Geocoding API to return.

lib/index.js
```diff
@@ -488,11 +482,29 @@ MapboxGeocoder.prototype = {
...
-        res.features = res.features
-          ? localGeocoderRes.concat(res.features)
-          : localGeocoderRes;
 
+        if (this.options.localGeocoder) {
+
+          localGeocoderRes = this.options.localGeocoder(searchInput, res.features) || [];
+
+          // supplement Mapbox Geocoding API results with features returned by a promise
+          if(typeof localGeocoderRes.then === 'function') {
```
andrewharvey, line 491: the JSDoc would need updating if we go with localGeocoder, but as mentioned, I have a slight lean towards leaving localGeocoder as is and adding a new externalGeocoder option. Open to your thoughts.

mapbox-danny:
@andrewharvey, your feedback to use externalGeocoder to add support for promises makes a lot of sense. I have updated the proposed approach to keep localGeocoder effectively unchanged and to add in the option for the externalGeocoder.

andrewharvey:
Oh wow you even managed to fix the broken unit tests unrelated to this feature. Thanks!

I haven't done a deep look at all the changes, but overall it looks good.

mapbox-danny:
> Oh wow you even managed to fix the broken unit tests unrelated to this feature. Thanks!
> 
> I haven't done a deep look at all the changes, but overall it looks good.

Thanks @andrewharvey. Let me know what you think about getting it ready to merge when you have a chance.

Issue #34, opened 2016-04-17
Capturing paste event is problematic on mac


fenec:
Would be good to catch a `paste` event on geocoder input

tmcw:
I'm not sure why this would be necessary, can you specify an example usecase for listening to `paste`? There are _many_ events that any input emits, and individually passing each event without a clear usecase will make this control's code very convoluted.

fenec:
@tmcw sure. the geocoder simply won't respond in any way if a user pastes address from clipboard. 

btw, examples seem to be broken: https://www.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/ 
I cannot see geocoder input

andrewharvey:
@fenec The examples are fixed now.

On Chrome, pasting with Ctrl+V or middle mouse (linux) will ensure the suggestion list is opened, however pasting with right click "Paste", doesn't open the suggestion list. Not sure if the `paste` event helps though, since I couldn't find the actual text contents in the paste event and the input is only filled in after the paste event is called.

andrewharvey:
fix proposed at https://github.com/tristen/suggestions/pull/17

edkim:
@andrewharvey Now that [tristen/suggestions#17](https://github.com/tristen/suggestions/pull/17) has been merged can we incorporate into the geocoder input?

On Mac, pasting into the geocoder search using CMD+V doesn't trigger the suggestions list. I have to press delete or another key to see a list of matching locations (CTRL works, but arrow keys don't).

andrewharvey:
@edkim not yet. We need to do a release of suggestions (https://github.com/tristen/suggestions/issues/19) with the latest changes so the dependency can be updated here.

andrewharvey:
@edkim I did a patch release of `suggestions` which included the bug fix for this. npm will pick up the latest patch release even without bumping the version here in package.json (not sure if we should bump it regardless). So simply redoing `npm install` followed by a rebuild should bring in those fixes.

Would be good to see a new release of `mapbox-gl-geocoder` to bring in the `suggestions` fixes and the `language` param.

ogdenstudios:
Hey everyone, this is still a bug as of v4.2.0. Using mapbox-gl-geocoder v4.2.0 with mapbox-gl-js v1.11.1. I had a handful of user bug reports that pasting addresses was not working. Verified it on OSX with Firefox 79. 

Had to fix it quick so I added a hack to fire off a phony `keydown` event: 

[...]

PR #390, opened 2020-08-30
where the supplied externalGeocoder method returns null, such as when after inspecting the search input it decides an external request is not needed, gracefully handle it by resolving with an empty results response

andrewharvey:

This issue was found by @ogdenstudios in https://github.com/mapbox/mapbox-gl-geocoder/issues/34#issuecomment-683362574

It is triggered by the demo, where after inspecting the search input it decides not to return a promise and instead returns null, in these cases it looks like the original code intended to handle this situation gracefully by returning an empty results, and I agree that should happen, however, because in the subsequent line at https://github.com/mapbox/mapbox-gl-geocoder/blob/a5591a3cab9e3d133aeff99d4f31a659f9d9d614/lib/index.js#L508 it tries to call .then, it must be a Promise not `[]`.

This should be done, but I haven't yet so leaving this PR in draft.

@mapbox-danny would you mind taking a look please?

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -502,8 +502,7 @@ MapboxGeocoder.prototype = {
             : localGeocoderRes;
 
           if (this.options.externalGeocoder) {
-
-            externalGeocoderRes = this.options.externalGeocoder(searchInput, res.features) || [];
+            externalGeocoderRes = this.options.externalGeocoder(searchInput, res.features) || Promise.resolve([]);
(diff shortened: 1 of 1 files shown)


mpothier:
This change was incorporated separately in PR [#446](https://github.com/mapbox/mapbox-gl-geocoder/pull/446), so I'm closing this one.

PR #394, opened 2020-09-01
Update README.md

malwoodsantoro:
Requires more coverage than a promise polyfill. See https://github.com/mapbox/mapbox-gl-js-docs/pull/399


andrewharvey (approved):
Thanks

PR #403, opened 2020-11-24
Fix window is not defined issue when doing server side rendering

SamSamskies:
This fixes the `window is not defined` error when doing server side rendering. The `suggestions` dependency is what was causing this issue and this issue has been fixed in v1.7.1. PR that fixed the issue https://github.com/tristen/suggestions/pull/42.


TheRedSpy15:
Any word on this getting merged? Seems silly that this PR has been collecting dust

JulienBlancher:
+1

PR #409, opened 2021-01-08
add russian placeholder

zavety:
Add Russian place holder in the search box.

### DIFF
--- a/lib/localization.js
+++ b/lib/localization.js
@@ -30,7 +30,8 @@ var placeholder = {
   'ko': '수색',//korean
   'pl':  'Szukaj', //polish
   'sl': 'Iskanje', //slovenian
-  'fa': 'جستجو' //persian(aka farsi)
+  'fa': 'جستجو',  //persian(aka farsi)
+  'ru': 'Поиск'//russian
 }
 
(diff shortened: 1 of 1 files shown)


andrewharvey (approved):
It doesn't look like the Mapbox Geocoding API supports Russian yet, so if you pass `ru` I'm not sure what would happen, but regardless I think it's okay to still localise the placeholder text in this way. But I'll leave this here a bit longer to let others weight in.

zavety:
Hello @andrewharvey , I've intergrated mapbox gl in my project, actually the api is able to return result in Russian. but there no Russian place holder in the search box. so I created this PR. thanks. also there's Russian support in the mapbox studio https://docs.mapbox.com/help/troubleshooting/change-language/

andrewharvey:
Yes, Mapbox Streets v8 map data supports Russian https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/ but I didn't see in Geocoding API docs Russian support https://docs.mapbox.com/api/search/geocoding/#language-coverage. But if you're seeing it, great. I guess I can merge this then.

PR #416, opened 2021-05-20
Fix flyTo coordinates not set when selected.center is not defined

thaddmt:
Currently when `selected.center` is not defined the `flyTo` action will just zoom in on the currently viewed area. Fixed this to be similar to how `selected.center` and `selected.geometry.coordinates` are handled in `_handleMarker` so that either can be used as the flyTo destination.

 - [x] write tests for all new functionality (no new functionality)
 - [x] run `npm run docs` and commit changes to API.md (no changes)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -375,7 +375,12 @@ MapboxGeocoder.prototype = {
           }
           flyOptions = extend({}, defaultFlyOptions, this.options.flyTo);
           //  ensure that center is not overriden by custom options
-          flyOptions.center = selected.center;
+          if (selected.center) {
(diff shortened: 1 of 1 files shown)


andrewharvey (approved):
Looks good to me.

thaddmt:
> Looks good to me.

Is there any way I can help to push this through? Looks like the travis CI build failed due to `not ok - Cannot create a client without an access token`. Is that a setting that I'm missing or is it in the travis CI build?

andrewharvey:
CI will fail for external contributors because the access token isn't set there, but should pass locally.

PR #418, opened 2021-05-31
Correct parameter list formating

luis-pato:
Show the "clear" event on the "on" method parameters list with the correct formating.

This is just a small change to the API README.
When i was going through the documentation o noticed the format on the list was a bit off, and i even missed that "clear" was also an event.
So i just made the small correction to the markup

### DIFF
--- a/API.md
+++ b/API.md
@@ -376,7 +376,8 @@ Subscribe to events that happen within the plugin.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


andrewharvey:
Thanks but unfortunately this is due to the upstream bug in documentationjs/documentation#676 prior PR's here #332 #261 #125 #166, the problem is they get replaced every time we re-generate the API docs. I don't think it's worth the effort manually patching this each time, so better to fix this either in documentation or the source file so that it get generated without this issue.

luis-pato:
Ah, ok, i'll just close this!

PR #419, opened 2021-06-08
Add config parameter to externalGeocoder call

thaddmt:
- Added the config object with geocoder options to the `externalGeocoder` function call so that the external geocoder can use options like proximity and language
- Found this because I've been trying to use `mapbox-gl-geocoder` with an externalGeocoder only like so:
- ```
  const externalGeocoderFn = async (input, _features, config) => {
    const { features } = await GeoAPI.forwardGeocode({
      query: input,
      proximity: config.proximity,
    });
    return features;
  };

    mapboxgl: maplibregl,
    localGeocoderOnly: true,
    localGeocoder: (input) => {},
    externalGeocoder: externalGeocoderFn,
  });
```
- Please let me know if there's a better way to do this or something I'm missing! Thanks!


Amplifiyer (approved):
Looks good to me.

PR #421, opened 2021-06-29
Added feature for showing multiple result markers when geocoding

thaddmt:
- Attempting to add a feature which shows the top query results as multiple markers on the map
- Added option `showResultMarkers` which gates this feature
- Ran into some issues getting this feature to work with the [suggestions](https://github.com/tristen/suggestions) library
  - `Suggestions` library seems to always have a selected item (defaults to first in the list) but I would like the multiple result markers to show up when a user presses the enter key with no items selected
  - This feature for now it is only triggered when enter key is pressed. So now when `showResultMarkers` is enabled you will not be able to mouse through results and press enter to fly to one of them. Enter key will always try to show multiple result markers
  - You can still use mouse clicks to fly to a specific item

Gif recording
Untitled


thaddmt:
Created a PR against (suggestions)[https://github.com/tristen/suggestions] which would make it possible to select no index value by default which would make this multiple result marker UX much better as i'll be able to move the multiple result marker code into the `_onChange` function instead of the `_onKeyDown` function

lib/index.js
```diff
@@ -1003,6 +1041,52 @@ MapboxGeocoder.prototype = {
...
+   * @returns {MapboxGeocoder} this
+   */
+  _handleResultMarkers: function(results){
+    // clean up any old marker that might be present
+    if (!this._map){
+      return;
+    }
+    this._removeResultMarkers();
+    var defaultMarkerOptions = {
+      color: '#4668F2'
```
Amplifiyer, line 1057: Is this default a standard?

Amplifiyer (approved):
Changes look good to me. Is _"when showResultMarkers is enabled you will not be able to mouse through results and press enter to fly to one of them"_ still valid?

lib/index.js
thaddmt: This is the same color used as the original `_handleMarker` method, just copied the color from there pretty much

thaddmt:
> Changes look good to me. Is _"when showResultMarkers is enabled you will not be able to mouse through results and press enter to fly to one of them"_ still valid?

That will be valid unless we switch off using `suggested` or this pr is accepted into `suggested` https://github.com/tristen/suggestions/pull/49

PR #422, opened 2021-06-30
Added a simple popup to markers which shows the place_name

thaddmt:
### Overview
- Added popup to marker and allows customizing popup options
- Popup sets text to `place_name` for now, open to changing how developer determines what to place in the popup


API.md
```diff
@@ -97,9 +97,11 @@ A geocoder component using the [Mapbox Geocoding API][55]
...
+    -   `options.popup` **([Boolean][61] \| [Object][56])** If `true`, a [Popup][68] will be added to the map when clicking on a marker using a default set of popup options.  If the value is an object, the popup will be constructed using these options. If `false`, no popup will be added to the map. Requires that `options.mapboxgl` also be set. (optional, default `true`)
     -   `options.render` **[Function][66]?** A function that specifies how the results should be rendered in the dropdown menu. This function should accepts a single [Carmen GeoJSON][67] object as input and return a string. Any HTML in the returned string will be rendered.
+    -   `options.popupRender` **[Function][66]?** A function that specifies how the results should be rendered in the popup menu. This function should accepts a single [Carmen GeoJSON][67] object as input and return a string. Any HTML in the returned string will be rendered.
```
Amplifiyer, line 102:
```suggestion
    -   `options.popupRender` **[Function][66]?** A function that specifies how the results should be rendered in the popup menu. This function should accept a single [Carmen GeoJSON][67] object as input and return a string. Any HTML in the returned string will be rendered.
```

lib/index.js
```diff
@@ -48,7 +48,9 @@ var subtag = require('subtag');
...
+ * @param {Boolean|Object} [options.popup=true]  If `true`, a [Popup](https://docs.mapbox.com/mapbox-gl-js/api/markers/#popup) will be added to the map when clicking on a marker using a default set of popup options.  If the value is an object, the popup will be constructed using these options. If `false`, no popup will be added to the map. Requires that `options.mapboxgl` also be set.
  * @param {Function} [options.render] A function that specifies how the results should be rendered in the dropdown menu. This function should accepts a single [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) object as input and return a string. Any HTML in the returned string will be rendered.
+ * @param {Function} [options.popupRender] A function that specifies how the results should be rendered in the popup menu. This function should accepts a single [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) object as input and return a string. Any HTML in the returned string will be rendered.
```
Amplifiyer, line 53:
```suggestion
 * @param {Function} [options.popupRender] A function that specifies how the results should be rendered in the popup menu. This function should accept a single [Carmen GeoJSON](https://github.com/mapbox/carmen/blob/master/carmen-geojson.md) object as input and return a string. Any HTML in the returned string will be rendered.
```

lib/index.js
```diff
@@ -88,7 +91,11 @@ MapboxGeocoder.prototype = {
     render: function(item) {
       var placeName = item.place_name.split(',');
       return '<div class="mapboxgl-ctrl-geocoder--suggestion"><div class="mapboxgl-ctrl-geocoder--suggestion-title">' + placeName[0]+ '</div><div class="mapboxgl-ctrl-geocoder--suggestion-address">' + placeName.splice(1, placeName.length).join(',') + '</div></div>';
-    }
+    },
+    popupRender: function(item) {
+      var placeName = item.place_name.split(',');
+      return '<div class="mapboxgl-ctrl-geocoder--suggestion popup-suggestion"><div class="mapboxgl-ctrl-geocoder--suggestion-title popup-suggestion-title">' + placeName[0] + '</div><div class="mapboxgl-ctrl-geocoder--suggestion-address popup-suggestion-address">' + placeName.splice(1, placeName.length).join(',') + '</div></div>'
```
Amplifiyer, line 97: Where do we define these classes `popup-suggestion`?

This seems very similar to the `render` method above, when and where would I choose this over the marker render?

Amplifiyer (commented):
Do we not need to worry about removing the popup for any cases?

lib/index.js
thaddmt: The render function above is used for rendering the search results list underneath the search bar. In this case I use a similar pattern to allow developers to override the popup's html while having a default method initially set

thaddmt:
> Do we not need to worry about removing the popup for any cases?

Can you elaborate what you mean by removing it @Amplifiyer?

lib/index.js
Amplifiyer: Got it, do we not need to define some default styling for `popup-suggestion`?

Amplifiyer:
> > Do we not need to worry about removing the popup for any cases?
> 
> Can you elaborate what you mean by removing it @Amplifiyer?

E.g. when user redo the search, don't we want to remove the popups from last search results?

thaddmt:
> > > Do we not need to worry about removing the popup for any cases?
> > 
> > 
> > Can you elaborate what you mean by removing it @Amplifiyer?
> 
> E.g. when user redo the search, don't we want to remove the popups from last search results?

The popup is removed when the marker is removed, when re-doing a search the markers are cleared out thus clearing out the popups as well!

lib/index.js
thaddmt: There are default styles defined on things like `mapboxgl-ctrl-geocoder--suggestion`, I added the `popup-suggestion` class name so that developers can override styles on the popup box only

Amplifiyer (approved):
Looks good to me.

Issue #408, opened 2020-12-29
Typing semicolon in the geocoder causes an error


jakub-leszczynski:
Hey!

The easiest way to reproduce is to:

1.  Navigate to geocoder example page here: https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/
2. Open the console to see error message
3. Type anything followed by a semicolon.
4. See the following message:
![Screenshot 2020-12-29 at 16 20 59 (2)](https://user-images.githubusercontent.com/26040789/103294531-514f5900-49f2-11eb-88ec-a12ea7e0b635.png)
![Screenshot 2020-12-29 at 16 20 46 (2)](https://user-images.githubusercontent.com/26040789/103294540-557b7680-49f2-11eb-9622-6f5230aa0ade.png)

I've been trying other signs but only the semicolon seems to be causing the issue.
I've also been able to reproduce this locally.

andrewharvey:
#352 was mostly for where a coordinate pair would create an error, but it could apply here. All we could do on the mapbox-gl-geocoder side is try and detect an error before the call is made and not make the call. Otherwise what else would you expect to happen here?

Issue #352, opened 2020-04-24
avoid making requests to the API which we know will return an error


andrewharvey:
cutting out an issue raised at https://github.com/mapbox/mapbox-gl-js-docs/issues/283 when users search for a pair of coordinates the api returns an error which is not caught.

I would call this a bug in mapbox-gl-geocoder, in that it shouldn't send a request to the API which it knows would return an error. So this means checking if the input looks like a coordinate pair and then doing something based on that (either drop the request or set limit).

The geocoder option reverseGeocode is not to false by default, to me that means if a user enters coordinates, don't do a reverse geocode to find nearby places, which I would say means just drop the query. If the requirement is that users can enter coordinates and have those coordinates appear as a search result then it needs to be handled like in https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-accept-coordinates/.

If the intent is to list nearby places, then reverseGeocode should be set to true and then it should list nearby places as results.

What's missing is if you want nearby places if coordinates entered but forward geocoding if not

cc @katydecorah

PR #424, opened 2021-07-09
Fix invalid geocoding requests

### LINKED ISSUE #274: geocoding/v5 API: MapiError 422 with coma query
Related to #168

Using Mapbox API geocoding/v5 with the `limit` param (`limit` option from *mapbox-gl-geocoder*), when you make a query with a coma whitout anything or with spaces after, such as:
```
42, 
```
Mapbox API returns a `422 MapiError` with 
```
{ message: limit must be combined with a single type parameter when reverse geocoding }
```
Working example:
https://codepen.io/ekqnp/pen/MdbrEx
This is also happening in the example proposed by Mapbox:
https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/

The only solution I found to avoid this error is to set `limit: false` and to handle the limit myself front-side.

Any ideas about it ?

karimnaaji:
This PR prevents forward geocoding requests to be interpreted as reverse geocoding requests.

Reminder on the accepted inputs from the server:
```
/geocoding/v5/{endpoint}/{longitude},{latitude}.json
```
```
/geocoding/v5/{endpoint}/{search_text}.json
```
Refer https://docs.mapbox.com/api/search/geocoding/#reverse-geocoding.

In the proposed PR, the geocoder follows this behavior:
- If `reverseGeocode=true`:
  - If the request is well-formed, it will be interpreted as a reverse geocoding request, and as a forward geocoding request otherwise.
- If `reverseGeocode=false`:
  - If the request is the start of a reverse geocoding request `number,number?`, or a valid geocoding request `number,number`, it will be transformed in order for the server to not interpret it as a reverse geocoding request (using comma trimming, as `12,` is equivalent to `12 ` when doing forward geocoding through this API). This effectively prevents 422 responses from the server for a user typing the start of an address with a number and a comma. 422 responses are the result of a forward geocoding request not being fully configured and falling through the `/{longitude},{latitude}.json` API syntax.

This also makes the option `reverseGeocode` intent more clear, meaning that no reverse geocode request may accidentally be interpreted as a forward geocoding request.

Fixes #274, #352, #408

### DIFF
--- a/API.md
+++ b/API.md
@@ -86,7 +86,7 @@ A geocoder component using the [Mapbox Geocoding API][55]
     -   `options.types` **[string][57]?** a comma seperated list of types that filter
         results to match those specified. See [https://docs.mapbox.com/api/search/#data-types][65]
         for available types.
-        If reverseGeocode is enabled, you should specify one type. If you configure more than one type, the first type will be used.
+        If reverseGeocode is enabled and no type is specified, the type defaults to POIs. Otherwise, if you configure more than one type, the first type will be used.
     -   `options.minLength` **[Number][60]** Minimum number of characters to enter before results are shown. (optional, default `2`)
     -   `options.limit` **[Number][60]** Maximum number of results to show. (optional, default `5`)
     -   `options.language` **[string][57]?** Specify the language to use for response text and query result weighting. Options are IETF language tags comprised of a mandatory ISO 639-1 language code and optionally one or more IETF subtags for country or script. More than one value can also be specified, separated by commas. Defaults to the browser's language settings.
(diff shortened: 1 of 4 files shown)


karimnaaji:
Tagging @rmrice and @andrewharvey for review as I see that you were involved in past changes there. Let me know if you have any questions on the change. We're doing a docs/example sprint and are trying to fix old issues, we've noticed tons of sentry exceptions triggered from these examples:
- https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder-accept-coordinates/
- https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-geocoder/

andrewharvey:
The new logic when `reverseGeocode=false` of transforming `number,number?` into something that will forward geocode without error is even better than what I had in mind (dropping the request). :+1:

rmrice (approved):
LGTM 🍦 

Verified that this works well locally 👍

Issue #401, opened 2020-11-20
setInput bug?


anatolysukhanov:
https://github.com/mapbox/mapbox-gl-geocoder/blob/master/API.md#setinput is used to only populate the input (at least 4.5.1).

So, for intstance, if I call

    geocoder.setInput("New York")

I get this:

![image](https://user-images.githubusercontent.com/7021042/99783416-74513800-2b2b-11eb-991c-f15bb1c377a0.png)

but in the latest version (4.7.0) it has changed and now it shows the results dropdown:

![image](https://user-images.githubusercontent.com/7021042/99783980-31dc2b00-2b2c-11eb-9569-2e0d593f73bb.png)

Is that a bug or intentional new behaviour?

zokioki:
This behavior was recently introduced via https://github.com/mapbox/mapbox-gl-geocoder/pull/345.

antony:
Just stumbled across this as well. It's hugely problematic if you are trying to prefill the location picker with the current value of the field. You now have only two choices to show:

1. A blank field
2. A dropdown open in the middle of your page

I'm certainly eager to see https://github.com/mapbox/mapbox-gl-geocoder/pull/425 merged in order to allow better control over this feature

nx:
I encountered the same problem. It's indeed a bit confusing that `setInput()` leads to new results being shown.

This did the trick for me. It's not ideal, but there doesn't seem to be a better solution.
```javascript
geocoder.current._inputEl.value = value;
```

christopherstyles:
This also works, but am looking forward to #425 as well 😄 

```javascript
const inputElements = document.getElementsByClassName(
  'mapboxgl-ctrl-geocoder--input',
);

if (inputElements.length > 0) {
  inputElements[0].value = "New York";
}
```

PR #425, opened 2021-07-10
Add option to display suggestion on setInput

victorneuret:
Add an optional parameter to choose whether to display the suggestion or not when using `setInput`

`setInput` used to not show the suggestions but is has been updated in https://github.com/mapbox/mapbox-gl-geocoder/pull/345 without a possibility to have the same behaviour than before. As stated by https://github.com/mapbox/mapbox-gl-geocoder/pull/345#issuecomment-804175282 it feels more like a breaking change than an improvement.

In this issue we can see the behaviour difference https://github.com/mapbox/mapbox-gl-geocoder/issues/401 .

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -939,10 +939,9 @@ MapboxGeocoder.prototype = {
     this._inputEl.value = searchInput;
     this._typeahead.selected = null;
     this._typeahead.clear();
-    if (showSuggestions && searchInput.length >= this.options.minLength) {
-      this._geocode(searchInput);
-    } else if (!showSuggestions) {
-      this._onChange();
+    if (searchInput.length >= this.options.minLength) {
+      showSuggestions ? this._geocode(searchInput) : this._onChange();
+    }
     }
     return this;
   },
--- a/lib/index.js
+++ b/lib/index.js
@@ -931,7 +931,10 @@ MapboxGeocoder.prototype = {
    * @param {boolean} [showSuggestions=false] display suggestion on setInput call
    * @returns {MapboxGeocoder} this
    */
-  setInput: function(searchInput, showSuggestions = true) {
+  setInput: function(searchInput, showSuggestions) {
+    if (showSuggestions === undefined) {
+      showSuggestions = false
+    }
     // Set input value to passed value and clear everything else.
     this._inputEl.value = searchInput;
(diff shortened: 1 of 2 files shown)


antony (approved):

lib/index.js
```diff
@@ -701,15 +701,21 @@ MapboxGeocoder.prototype = {
...
+    }
     // Set input value to passed value and clear everything else.
     this._inputEl.value = searchInput;
     this._typeahead.selected = null;
     this._typeahead.clear();
-    if (searchInput.length >= this.options.minLength) {
+    if (showSuggestions && searchInput.length >= this.options.minLength) {
       this._geocode(searchInput);
+    } else if (!showSuggestions) {
+      this._onChange();
```
antony, line 718:
```suggestion
    if (searchInput.length >= this.options.minLength) {
      showSuggestions ? this._geocode(searchInput) : this._onChange();
    }
```

lib/index.js
```diff
@@ -701,15 +701,21 @@ MapboxGeocoder.prototype = {
...
    * Set input
    * @param {string} searchInput location name or other search input
+   * @param {boolean} [showSuggestions=false] display suggestion on setInput call
    * @returns {MapboxGeocoder} this
    */
-  setInput: function(searchInput) {
+  setInput: function(searchInput, showSuggestions) {
+    if (showSuggestions === undefined) {
+      showSuggestions = false
+    }
```
antony, line 710:
```suggestion
  setInput: function(searchInput, showSuggestions = false) {
```

antony (changes requested):
Just some minor code style changes. I wonder if there's a way we can add a test for this too?

antony:
When testing my suggestions I noticed that it will need an upgrade of uglify from v2 to v3, since uglify v2 is too old to be able to handle default parameters.

If this isn't acceptable we could drop the default parameter and go back to the `=== undefined` logic.

lib/index.js
victorneuret: I tried this at first but it is not possible due to uglify version as you mentioned bellow
I don't think that upgrading the version in this PR is useful.

rewathkafley (approved):

rewathkafley (approved):

PR #434, opened 2021-09-15
Added autocomplete, fuzzyMatch, routing, worldview parameters

mpothier:
- added these missing params to geocoder options
- public getter and setter method for each
- tests for each new parameter
- bumped mapbox-sdk-js to latest version to enable handling these missing params

 - [X] briefly describe the changes in this PR
 - [X] write tests for all new functionality

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1083,7 +1083,7 @@ MapboxGeocoder.prototype = {
 
   /**
    * Get the current worldview parameter value used for requests
-   * @returns {Boolean} The worldview parameter value
+   * @returns {String} The worldview parameter value
    */
   getWorldview: function(){
     return this.options.worldview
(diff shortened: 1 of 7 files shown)


lib/index.js
```diff
@@ -1006,6 +1017,78 @@ MapboxGeocoder.prototype = {
...
+   * @returns 
+   */
+  setWorldview: function(code){
+    this.options.worldview = code;
+    return this;
+  },
+
+  /**
+   * Get the current worldview parameter value used for requests
+   * @returns {Boolean} The worldview parameter value
```
mapbox-danny, line 1086: Update worldview response to @returns {String}

mapbox-danny (approved):

lib/index.js
mpothier: Good catch! Done

Issue #185, opened 2019-02-20
Debounce option


matthewlein:
Hello,

I am using a custom endpoint that calls, and I could use an option to set the debounce ms as an option, to try to reduce the load on that endpoint.

I'm happy to make the PR, it seems like it will be just 2 lines and documentation.

https://github.com/mapbox/mapbox-gl-geocoder/blob/b69197a6852d7ca6337e9efef9e687efab70b786/lib/index.js#L58-L67
https://github.com/mapbox/mapbox-gl-geocoder/blob/b69197a6852d7ca6337e9efef9e687efab70b786/lib/index.js#L164

andrewharvey:
Personally I don't mind that being configurable, however recently I've been thinking if this would be better to be throttle rather than debounce.

My reasoning is, if I keep on typing within the 200ms debounce, the request will never be sent until I finish typing. As a user I keep on typing since I haven't seen any suggestions yet.

Whereas using throttle, at least I'll start seeing some results show up, even if they are outdated by a few chars, I'd rather see some suggestions start to popup up instead of none.

The reason I bring this up is if we add a public debounceWait option, but then swap to throttle internally it creates a breaking API change, so best if we can decide on this first. I'll break this out into a separate ticket.

matthewlein:
Yep, I agree with that UX, although it would end up hitting my server even more, which is what I'm trying to control. In that regard, choosing between debounce or throttle, and also setting the timeout is the most useful option.

PR #436, opened 2021-10-12
Configurable debounce delay

kvalev:
Allow the users to configure the debounce delay, which might be helpful when querying an `externalGeocoder`. 


wibed:
> Thanks for the PR. My view is per #185 we shouldn't be using debounce at all and instead use throttle. I'd like to see more thought and discussion and ultimately a decision on this point before we go adding to the API. If we add this to the API now then swap to throttle it would be a breaking change which is something we should try to avoid.

throttling would fire the first event and then wait for a given time until the next event fires.
debounce would cancel events until a threshold has been met.

a debounce would be helpful to make the first event as late as possible, while throttle would prevent the app to fire repeatedly within a given period.

why not both?

JoshMLeslie:
So hey what's up? Been a few years. Would love to have the option to change debounce time or even submit my own function into the mix so I'm not chewing up API calls with random characters.

PR #446, opened 2021-12-20
fix unsafe HTML construction with createIcon

stepankuzmin:
This PR removes IE support for icon creation and removes `createIcon` from the `MapboxGeocoder` prototype.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -276,16 +276,7 @@ MapboxGeocoder.prototype = {
     icon.setAttribute('xml:space','preserve');
     icon.setAttribute('width', 18);
     icon.setAttribute('height', 18);
-    // IE does not have innerHTML for SVG nodes
-    if (!('innerHTML' in icon)) {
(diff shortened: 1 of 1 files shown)


ryanhamley (commented):
This looks good. I understand the rationale for removing `createIcon` from `index.js`, but this does technically change the API. It's not a documented method, but I think removing it might still necessitate a major version change, so maybe it's better to leave it be.

stepankuzmin:
Thanks Ryan!

As per discussion with @mapbox-danny we can hold on the merge until a major version update

ryanhamley:
@stepankuzmin I think it's ok to revert the change that removes `createIcon` from the public API and just remove the offending IE11 code. That will resolve the open security ticket.

stepankuzmin:
@ryanhamley sure, I've updated the PR

mapbox-danny (approved):
Thanks for adding the Promise.resolve fix too

PR #447, opened 2022-01-23
Add babelify to build & update nanoid to fix security warning

### LINKED ISSUE #443: Uncaught TypeError: nanoid is not a function
I'm getting

Uncaught TypeError: nanoid is not a function
    at MapboxEventManager.generateSessionID

Looking at the code, and some other reports of issues like this, I suspect the code in events.js should be doing:

```
var { nanoid } = require('nanoid')
```

But then I don't understand why this works in the mapbox examples??

justisb:
## Summary

- Introduces Babel/Babelify to the publish script, to ensure output retains ES5-compatibility- the update to `nanoid` would have otherwise introduced object destructuring not present in the dist previously, and would be incompatible with the non-harmony UglifyJS used by this package. Another option could be to switch to `uglify-es` and eschew ES5 compatibility.
- Bumps `nanoid` from 2.1.11 to 3.1.31 to resolve security vulnerability warning for [CVE-2021-23566](https://github.com/advisories/GHSA-qrpm-p2h7-hrv2), present in versions prior to 3.1.31- I don't think this package's use-case of nanoid is a security issue in itself, but it's nice to have the dependency updated to satisfy Dependabot/Snyk, and to de-dupe it with other packages that keep their dependencies updated.
- It should be easier to update other dependencies now as well.

`npm run prepbulish` succeeds & `npm test` passes for me locally- not sure if there is something else I need to look at to fix the Travis CI build.

Likely fixes #443.

## PR Checklist

 - [x] ~write tests for all new functionality~
 - [x] ~run `npm run docs` and commit changes to API.md~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## master
 
-### Dependency Update
+### Dependency update
 
-- Bumps `nanoid` to v3.1.31 to resolve security vulnerability warning
+- Bumps `nanoid` to v3.1.31 to resolve security vulnerability warning.
+- Adds `babelify` to build process to ensure mapbox-gl-geocoder remains ES5-compatible.
 
 ## 4.7.4
 
@@ -13,7 +14,7 @@
 
 ## 4.7.3
 
-### Dependency Update
+### Dependency update
 
 - Bumps `mapbox-sdk-js` to v0.13.1 to resolve issue with `got` library.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## master
 
+### Dependency Update
+
+- Bumps `nanoid` to v3.1.31 to resolve security vulnerability warning
+
 ## 4.7.4
 
 ### Features / Improvements 🚀
@@ -9,7 +13,7 @@
 
 ## 4.7.3
 
-### Dependency update
+### Dependency Update
 
 - Bumps `mapbox-sdk-js` to v0.13.1 to resolve issue with `got` library.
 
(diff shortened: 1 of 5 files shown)


justisb:
After seeing the prepublish script failure, ended up introducing perhaps a more controversial change- because the newer version of `nanoid` contains ES6 syntax, it required also upgrading to the ES6-compatible version of UglifyJS.

Not sure if that's the preferred solution, or if we would want to introduce Babel to run against dependencies first in the build script, to ensure this package's dist output does not contain ES6 features.

justisb:
Switched this back to a draft PR while I mess around with a Babel/Babelify config, so that there's no compromise in ES5 compatibility.

justisb:
Updated the PR description to explain the rationale, though open to other solutions.

Also not sure if the build failure of `throw new Error('Cannot create a client without an access token')` is due to these changes- it seems unrelated?

mbullington (approved):
Browserify is a name I haven't heard in a very long time, ha. Thank you for your contributions!

I think given the wide base of support GL Geocoder has, ES5 compatibility isn't something we feel like disrupting. Babelify is a good call and thank you for setting it up.

Looks good on my end, I'll play around with getting the builds to work.

mbullington:
Seems like if I make the branch tests work fine, a security issue allowing external contributors access to our access tokens.

Thanks again!

justisb:
> Browserify is a name I haven't heard in a very long time, ha. Thank you for your contributions!

For sure- I did consider switching to some more modern dependencies, but thought it'd be safer to have minimal effect on the published version's compatibility. Which hopefully works- do you know what the release timing usually looks like? Thanks for the quick merge!

PR #449, opened 2022-01-28
Add method to refresh accessToken

tristen:
Adds a method to update the accessToken after the Geocoder has been initialized. This is helpful when you wish to revise a token within a long-running application and the session of a token is shorter.


mbullington:
image

I'll make sure to review in a lil

mbullington (approved):
Looks pretty straightforward. Thanks!

PR #485, opened 2023-01-17
useGeocoderV6 flag for MapboxGeocoder

VladislavMedved:
## Description

PR adds `useGeocoderV6` flag to the options of `MapboxGeocoder` class that allows using v6 endpoint instead of default v5 when geocoding.

It introduces a new mode for forward geocoding called `structured` for increasing the accuracy of results.

The full documentation for v6 API [here](https://docs.mapbox.com/api/search/geocoding-v6/)


lib/index.js
```diff
@@ -90,6 +92,7 @@ function MapboxGeocoder(options) {
   this.fresh = true;
   this.lastSelected = null;
   this.geolocation = new Geolocation();
+  return this.options.useGeocoderV6 ? new MapboxGeocoderV6(options) : this;
```
VladislavMedved, line 95: @mpothier  Do we want to use v6 functionality via  useGeocoderV6 flag in scope of the old MapboxGeocoder, or we should consider exposing another class MapboxGeocoderV6?

lib/index.js
```diff
@@ -75,6 +76,7 @@ function getFooterNode() {
  * @param {Boolean} [options.routing=false] Specify whether to request additional metadata about the recommended navigation destination corresponding to the feature or not. Only applicable for address features.
```
VladislavMedved, line 76: @mpothier set of parameters could be different. For example we have `routing` and `fuzzyMatch` in v5, and have `permanent` parameter in v6

lib/utils.js
```diff
@@ -60,6 +66,34 @@ function getAddressInfo(feature) {
...
+/**
+ * This function transforms the v6 feature from reverse geocoding to AddressInfo object
+ * @param {object} feature 
+ * @returns {object}
+ */
+function getV6AddressInfo(feature) {
+  const houseNumber = feature.address_number || "";
+  const street = feature.street || "";
+  const placeName = feature.place_formatted || "";
+  const address = placeName.split(",")[0];
```
mpothier, line 78: Use `feature.name`

lib/utils.js
```diff
@@ -60,6 +66,34 @@ function getAddressInfo(feature) {
...
 
+/**
+ * This function transforms the v6 feature from reverse geocoding to AddressInfo object
+ * @param {object} feature 
+ * @returns {object}
+ */
+function getV6AddressInfo(feature) {
+  const houseNumber = feature.address_number || "";
+  const street = feature.street || "";
+  const placeName = feature.place_formatted || "";
```
mpothier, line 77: Use
```js
feature.name + (feature.place_formatted ? `, ${feature.place_formatted}` : "")
```

lib/utils.js
```diff
@@ -60,6 +66,34 @@ function getAddressInfo(feature) {
...
+    address: address,
+    houseNumber: houseNumber,
+    street: street,
+    placeName: placeName,
+  };
+
+  for (const contextProp in feature.context) {
+    if (contextProp) {
+      const layer = contextProp.id.split(".")[0];
+      addrInfo[layer] = contextProp.text;
```
mpothier, line 90: Update for v6 schema:
```js
addrInfo[contextProp] = feature.context[contextProp].name
```

lib/index.js
```diff
@@ -840,7 +898,7 @@ MapboxGeocoder.prototype = {
     if (!results.features.length) return;
     var result = results.features[0];
     this._typeahead.selected = result;
-    this._inputEl.value = result.place_name;
+    this._inputEl.value = this.options.version === 'v6' ? result.properties.name : result.place_name;
```
mpothier, line 901: We need to concatenate `result.properties.name` and `result.properties.place_formatted` for V6 features

lib/index.js
```diff
@@ -653,10 +695,26 @@ MapboxGeocoder.prototype = {
       if (reverseGeocodeCoordRgx.test(trimmedSearch)) {
         search = search.replace(/,/g, ' ');
       }
+      if(this.options.version === 'v6' && config.mode) {
+        config.permanent = config.mode === 'mapbox.places-permanent';
+        delete config.mode;
+      }
```
mpothier, line 701: This should be moved outside of the switch block, as it could apply to forward or reverse.

lib/index.js
```diff
@@ -83,13 +85,47 @@ function getFooterNode() {
...
+    ? `<div class="mapboxgl-ctrl-geocoder--suggestion-title">${name}</div>`
+    : ""
+}
+      ${
+  place_formatted
+    ? `<div class="mapboxgl-ctrl-geocoder--suggestion-address">${place_formatted}</div>`
+    : ""
+}
+    </div>`;
+}
```
mpothier, line 116: For organizational and IDE purposes, let's move these functions after `function MapboxGeocoder(options) {...}` and before `MapboxGeocoder.prototypes = {...}`. The main reason is that, in my IDE (VS Code), the JSDoc string for the MapboxGeocoder class is showing up on the highlight tooltip for the first function after the doc string, which in this case makes it look like `getItemValue` is a class instead of a function:
image

debug/index.js
```diff
@@ -85,7 +85,24 @@ var geocoder = new MapboxGeocoder({
...
+  searchType: 'structured',
+  queryCoordinates: {longitude: 40.76490756883251, latitude: -74.0437572064368},
+  address_line1: 'address_line1',
+  address_number: 'address_number',
+  street: 'street',
+  place: 'place',
+  region: 'region',
+  countries: 'US',
+  postcode: 'postcode',
+  locality: 'locality',
```
mpothier, line 98: Let's remove these if they aren't used

debug/index.js
```diff
@@ -85,7 +85,24 @@ var geocoder = new MapboxGeocoder({
...
+
+  version: 'v6',
+  types: "address",
+  mode: 'mapbox.places-permanent',
+  reverseGeocode: true,
+  flipCoordinates: true,
+  permanent: true
```
mpothier, line 105: Let's remove this if it's not used (since `mode` is used instead)

mpothier (changes requested):
@VladislavMedved Great work, this is really close - just a few small changes

lib/index.js
VladislavMedved: nice catch, but it should be moved before MapboxGeocoder class, just also before the JSDoc annotation, cause eslint will give you error if you use something before it's declaration

lib/index.js
VladislavMedved: We use only name as an input value for search control

Screen Shot 2023-01-26 at 2 31 58 PM

lib/utils.js
```diff
@@ -60,6 +66,34 @@ function getAddressInfo(feature) {
...
+    address: address,
+    houseNumber: houseNumber,
+    street: street,
+    placeName: placeName,
+  };
+
+  for (const contextProp in feature.context) {
+    if (contextProp) {
+      const layer = contextProp.id.split(".")[0];
+      addrInfo[layer] = contextProp.name;
```
mpothier, line 90: I believe you have you index the `feature.context` object with the `contextProp` key:
```js
addrInfo[contextProp] = feature.context[contextProp].name
```
otherwise `contextProp` only refers to the key rather than the full object

lib/index.js
mpothier: This is specific to the `query` method that allows programmatic triggering of a search (searches for a string and automatically "selects" the first result). The existing implementation uses the full `place_name` string for v5.

mpothier (changes requested):

debug/index.js
```diff
@@ -85,7 +85,10 @@ var geocoder = new MapboxGeocoder({
         .then(response => response.json())
     }
   },
-  mapboxgl: mapboxgl
+  mapboxgl: mapboxgl,
+  reverseGeocode: true,
+  flipCoordinates: true,
```
mapbox-danny, line 90: Do we have flip-coordinates on by default in v6? Let's discuss

lib/index.js
```diff
@@ -31,6 +32,36 @@ function getFooterNode() {
   return div;
 }
 
+function getItemValue(item) {
+  return item.place_name
+}
+
+function render(item) {
+  var placeName = item.place_name.split(',');
+  return '<div class="mapboxgl-ctrl-geocoder--suggestion"><div class="mapboxgl-ctrl-geocoder--suggestion-title">' + placeName[0]+ '</div><div class="mapboxgl-ctrl-geocoder--suggestion-address">' + placeName.splice(1, placeName.length).join(',') + '</div></div>';
```
mapbox-danny, line 41: Let's use the place_formatted from the v6 response object

lib/index.js
```diff
@@ -31,6 +32,36 @@ function getFooterNode() {
...
+}
+
+function renderV6(item) {
+  const { name, place_formatted } = item.properties;
+  return `<div class="mapboxgl-ctrl-geocoder--suggestion">
+      ${
+  name
+    ? `<div class="mapboxgl-ctrl-geocoder--suggestion-title">${name}</div>`
+    : ""
+}
```
mapbox-danny, line 56: It looks like the whitespace formatting may be off here.

lib/index.js
```diff
@@ -75,6 +106,7 @@ function getFooterNode() {
  * @param {Boolean} [options.routing=false] Specify whether to request additional metadata about the recommended navigation destination corresponding to the feature or not. Only applicable for address features.
  * @param {String} [options.worldview="us"] Filter results to geographic features whose characteristics are defined differently by audiences belonging to various regional, cultural, or political groups.
  * @param {Boolean} [options.enableGeolocation=false] If `true` enable user geolocation feature.
+ * @param {'v5'|'v6'} [options.version='v5'] If `v6` use v6 endpoint instead of v5. The default value is set to v5 API.
```
mapbox-danny, line 109: Do we want to make the default v6 going forward?

lib/index.js
```diff
@@ -110,13 +146,7 @@ MapboxGeocoder.prototype = {
...
     enableGeolocation: false,
     addressAccuracy: 'street',
-    getItemValue: function(item) {
-      return item.place_name
-    },
-    render: function(item) {
-      var placeName = item.place_name.split(',');
-      return '<div class="mapboxgl-ctrl-geocoder--suggestion"><div class="mapboxgl-ctrl-geocoder--suggestion-title">' + placeName[0]+ '</div><div class="mapboxgl-ctrl-geocoder--suggestion-address">' + placeName.splice(1, placeName.length).join(',') + '</div></div>';
-    }
+    version: 'v5'
```
mapbox-danny, line 149: Confirm this updates with  "version" change

mapbox-danny:
Be sure to remove the yarn lock

debug/index.js
VladislavMedved: No, it's debug file for testing purposes. flip coordinates by default set to false
https://github.com/mapbox/mapbox-gl-geocoder/blob/d49005774eff3bd0fd3edc8b4e521b4b9a5cc9bb/lib/index.js#L138

lib/index.js
VladislavMedved: I would not use it as default value because it breaks backward compatibility. Because v6 contact is very different from v5 one, every user that expects to receive v5 suggestion will get their code broken

lib/index.js
VladislavMedved: could you specify what do you mean?

VladislavMedved:
- [ ] Update the version number in `package.json` and `package-lock.json`
- [ ] Run npm run docs and commit changes to API.md
- [ ] Update CHANGELOG.md with changes under main heading before merging

MateuszKukiela:
Is anyone still working on this? It would be really cool to see API v6 in the geocoder

MateuszKukiela:
Pinging once again because it's getting ridiculous.

mpothier:
@MateuszKukiela I recommend checking out the Mapbox Search JS SDK, which is the newer SDK and uses the Geocoding V6 API in its libraries:
- https://docs.mapbox.com/mapbox-search-js/example/add-geocoder-to-map/
- https://docs.mapbox.com/mapbox-search-js/api/web/geocoding/

PR #502, opened 2023-07-05
Fix setPlaceholder to update this.options.placeholder

marexandre:
Fix `setPlaceholder` so that when `setPlaceholder` is called afterwards it return updated placeholder string

### Before
Example of the bug and what this PR tries to fix:

```js
    accessToken: mapboxgl.accessToken,
    mapboxgl: mapboxgl,
    placeholder: 'Test'
});

geocoder.getPlaceholder();
// returns Test"

geocoder.setPlaceholder('New string');

geocoder.getPlaceholder();
// returns  "Test"  and not the updated placeholder text
```

### After

```js
    accessToken: mapboxgl.accessToken,
    mapboxgl: mapboxgl,
    placeholder: 'Test'
});

geocoder.getPlaceholder();
// returns  "Test"

geocoder.setPlaceholder('New string');

geocoder.getPlaceholder();
// returns  "New string" 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## HEAD
 
+## 5.0.2
+
+### Bug fixes 🐛
+
+- Fix `setPlaceholder` so that when `setPlaceholder` is called afterwards it return updated placeholder string [#502](https://github.com/mapbox/mapbox-gl-geocoder/pull/502)
+
 ## 5.0.1
 
 ### Bug fixes 🐛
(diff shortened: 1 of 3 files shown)


marexandre:
If I wanted to have this change in `v4` of this library, what would be the recommendations to add a similar fix?

marexandre:
CI is failing for the following reason - which does not seem to be related to changes in this PR...

Screen Shot 2023-07-05 at 12 03 23 PM

stepankuzmin (approved):
Hi @marexandre,

Thanks for the contribution, overall everything looks good to me 👍
Unfortunately, CI will fail for external contributors because the access token isn't set there. But it should pass locally when you add the `MapboxAccessToken` environment variable.

marexandre:
@stepankuzmin thank you for taking a look at this changes - rant the tests locally and all passed 🎉 
How should we proceed with merging this PR? Anything else I should do or missed?

PR #529, opened 2024-09-11
Update event schema

mpothier:
## Description

- Updates the event schema to version 2.2 for `keystroke` and `select` events. Added some validation to payload constructions to ensure server compatibility.
- Enable logging with filters and localGeocoder results mixed in. This is made possible by tagging a `_source` for Mapbox results to distinguish from client-added results.

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -386,7 +386,12 @@ MapboxEventManager.prototype = {
    * @returns 
    */
   objectHasRequiredProps: function(obj, requiredProps) {
-    return requiredProps.every(prop => obj[prop] !== undefined);
+    return requiredProps.every(function(prop) {
+      if (prop === 'queryString') {
+        return typeof obj[prop] === 'string' && obj[prop].length > 0;
+      }
+      return obj[prop] !== undefined;
+    });
   },
 
   /**
--- a/lib/events.js
+++ b/lib/events.js
@@ -397,10 +397,6 @@ MapboxEventManager.prototype = {
   shouldEnableLogging: function(options){
     if (options.enableEventLogging === false) return false;
     if (options.origin && options.origin !== 'https://api.mapbox.com') return false;
-    // hard to make sense of events when a local instance is suplementing results from origin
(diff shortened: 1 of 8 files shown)


lib/events.js
```diff
@@ -146,24 +141,41 @@ MapboxEventManager.prototype = {
...
     // Handle proximity, whether null, lat/lng coordinate object, or 'ip'
     var proximity;
     if (!geocoder.options.proximity) {
       proximity = null;
     } else if (typeof geocoder.options.proximity === 'object') {
       proximity = [geocoder.options.proximity.longitude, geocoder.options.proximity.latitude];
     } else if (geocoder.options.proximity === 'ip') {
-      proximity = [999,999];  // Alias for 'ip' in event logs
+      var ipProximityHeader = geocoder._headers ? geocoder._headers['ip-proximity'] : null;
+      if (ipProximityHeader) {
```
mmarkov, line 166: nit; Do you think it worth to check for `typeof ipProximityHeader === 'string'` to be sure function `split` is availble.

lib/index.js
```diff
@@ -118,6 +118,8 @@ MapboxGeocoder.prototype = {
       return '<div class="mapboxgl-ctrl-geocoder--suggestion"><div class="mapboxgl-ctrl-geocoder--suggestion-title">' + placeName[0]+ '</div><div class="mapboxgl-ctrl-geocoder--suggestion-address">' + placeName.splice(1, placeName.length).join(',') + '</div></div>';
     }
   },
+  
+  _headers: {},
```
mmarkov, line 122: nit; empty object is "truthy", maybe worth to put `null` as initial value?

mmarkov (approved):
lgtm;

lib/events.js
mpothier: Good idea

lib/index.js
mpothier: I think this is ok for how we are using it.

PR #559, opened 2026-08-05
RTFDB-4842: Improve user input handling in "mapbox-gl-geocoder"

pahuta:
- Add function to parse different spatial formats which can be used to configure `localGeocoder` function
- Add `isValidTile`, `isValidQuadkey`, `tileToLngLat` and `quadkeyToTile` helpers to `lib/utils.js`
- Use the parsed zoom when flying to a selected feature that has no `bbox`
- Fix the suggestion list not updating after cutting (Cmd/Ctrl+X) or undoing (Cmd/Ctrl+Z) a change to the search input

https://mapbox.atlassian.net/browse/RTFDB-4842

 - [x] update CHANGELOG.md with changes under `main` heading before merging

### Parsing extended spatial formats when enabled:

https://github.com/user-attachments/assets/03f8be83-2d76-4559-adfe-50ca20787b83

### Previous vs current behavior when user cut or undo search input value

https://github.com/user-attachments/assets/b1f2340e-5a87-4bf2-acfe-03e77b92087c

### DIFF
--- a/debug/index.js
+++ b/debug/index.js
@@ -17,6 +17,7 @@ insertCss(
 );
 
 var MapboxGeocoder = require('../');
+var spatialFormats = require('../lib/spatial-formats');
 
 var mapDiv = document.body.appendChild(document.createElement('div'));
 mapDiv.style.position = 'absolute';
@@ -75,14 +76,15 @@ var geocoder = new MapboxGeocoder({
   trackProximity: true,
   useBrowserFocus: true,
   enableGeolocation: true,
-  parseExtendedSpatialFormats: {
-    commaSeparatedLngLatZoom: true,
-    slashSeparatedZoomLatLng: true,
-    tile: true,
-    quadkey: true
-  },
   localGeocoder: function(query) {
-    return coordinatesGeocoder(query);
+    var spatialFeatures = [
+      spatialFormats.parseSlashSeparatedZoomLatLng(query),
+      spatialFormats.parseCommaSeparatedLngLatZoom(query),
+      spatialFormats.parseTile(query),
+      spatialFormats.parseQuadkey(query)
+    ].filter(Boolean);
+
+    return spatialFeatures.concat(coordinatesGeocoder(query) || []);
   },
   externalGeocoder: function(query, features) {
     // peak at the query and features before calling the external api
--- a/debug/index.js
+++ b/debug/index.js
@@ -75,6 +75,12 @@ var geocoder = new MapboxGeocoder({
   trackProximity: true,
   useBrowserFocus: true,
   enableGeolocation: true,
+  parseExtendedSpatialFormats: {
+    commaSeparatedLngLatZoom: true,
+    slashSeparatedZoomLatLng: true,
+    tile: true,
+    quadkey: true
+  },
   localGeocoder: function(query) {
     return coordinatesGeocoder(query);
   },
--- a/lib/index.js
+++ b/lib/index.js
@@ -12,7 +12,6 @@ var localization = require('./localization');
 var subtag = require('subtag');
 var Geolocation = require('./geolocation');
 var utils = require('./utils');
-var spatialFormats = require('./spatial-formats');
 
 
 const GEOCODE_REQUEST_TYPE = {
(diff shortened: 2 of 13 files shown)


debug/index.js
```diff
@@ -75,6 +75,12 @@ var geocoder = new MapboxGeocoder({
   trackProximity: true,
   useBrowserFocus: true,
   enableGeolocation: true,
+  parseExtendedSpatialFormats: {
```
stepankuzmin: Why do we need to expose this as a public API?

lib/spatial-formats.js
```diff
@@ -0,0 +1,219 @@
+'use strict';
+
+var utils = require('./utils');
+
+// Marks a feature synthesized from an extended spatial format, so it can be told
+// apart from Geocoding API results (which carry `_source: 'mapbox'`).
+const SOURCE = 'extended-spatial-format';
```
stepankuzmin: Why do we need syntetic features at all?

debug/index.js
pahuta: To give our users (including internal ones) the ability to quickly navigate to a desired location. It's practically the same as how coordinate search works now.

stepankuzmin: So you expect end users to configure the accepted formats in UI?

lib/spatial-formats.js
pahuta: The same reason as in previous comment. I agree that it's not directly related to the geocoding but it significantly improve UX. For example in our user facing applications in Mapbox Console user will be able to quickly move the map to the desired location if he/she has coordinates in other format than apps support.

stepankuzmin: The feature itself is good, but I wonder if we could simplify the implementation

debug/index.js
pahuta: No, not on UI. Let's clarify, in this context "end users" are developers and they will just enable parsing **predefined** formats. Here is an example how we configure `mapbox-gl-geocoder` right now:
```ts
      accessToken,
      origin: config.MAPBOX_API_URL,
      useBrowserFocus: true,
      mapboxgl: mapboxgl,
      marker: false,
      reverseGeocode: true,
      flipCoordinates: true
    });
```

These are just a few additional configurations which are fully backwards compatible with a current default behavior. Example how new `parseExtendedSpatialFormats` will be used:

```ts
      accessToken,
      origin: config.MAPBOX_API_URL,
      useBrowserFocus: true,
      mapboxgl: mapboxgl,
      marker: false,
      reverseGeocode: true,
      flipCoordinates: true,
      // new config below
      parseExtendedSpatialFormats: {
        commaSeparatedLngLatZoom: true,
        slashSeparatedZoomLatLng: true,
        tile: true,
        quadkey: true
      }
    });
```

After that users on UI will be able to specify coordinates/tile/quadkey in the search input and navigate to the location.

lib/spatial-formats.js
pahuta: I open to the suggestion. In general code of `mapbox-gl-geocoder` is old and written in "all in one file" style. I tried to keep it maintainable by putting new functionality separately as much as possible and maintaining backward compatibility.

debug/index.js
stepankuzmin: I still don't understand why this needs to be configurable. Do you expect developers to explicitly opt out of some formats? Also, why did you choose these formats (i.e., why lnglat and not latlng)? I wonder if we could enable this by default, improve the detection logic, and avoid introducing such a public API at all.

pahuta: Changed the implementation as we discussed: indeed `localGeocoder` covers our needs and there is no sense to introduce the `parseExtendedSpatialFormats` parameter.

lib/index.js
```diff
@@ -968,18 +985,22 @@ MapboxGeocoder.prototype = {
           this._hideLoadingIcon();
           this._hideAttribution();
 
-          // in the event of an error in the Mapbox Geocoding API still display results from the localGeocoder
-          if ((localGeocoderRes.length && this.options.localGeocoder) || (externalGeocoderRes.length && this.options.externalGeocoder) ) {
```
stepankuzmin, line 972: Doesn't that break the `externalGeocoder`?

pahuta: No - that branch was already dead code for `externalGeocoder` before the removal. At that point `externalGeocoderRes`:

- holds an empty array (initial value) if the geocoding request failed the very first time
- holds a Promise (from `this.options.externalGeocoder(...) || Promise.resolve([])`), not a resolved array, so `externalGeocoderRes.length` was always undefined/falsy.

The condition could never be satisfied through the `externalGeocoder` side. So removing it doesn't change behavior for `externalGeocoder` at all.

